PR #134, opened 2019-03-27
✨ Service token auth

### LINKED ISSUE #110: Add support for service tokens
Other services within kids first will need to interact with the study creator. This means that the study creator should accept service tokens from ego, which are distinct from the currently supported user tokens.
The first pass at this may be as easy as checking if a token is a service token, then giving the request's user the `ADMIN` role so that the request will be treated exactly like an admin's request.

dankolbman:
Authenticates tokens with `client-credentials` grant type and correct scope as an ADMIN user.
We may wish to revisit this if we refactor our auth to make better use of permissions.

Closes #110

### DIFF
--- a/creator/middleware.py
+++ b/creator/middleware.py
@@ -169,6 +169,14 @@ class Auth0AuthenticationMiddleware():
         # Currently unused
         permissions = token.get('https://kidsfirstdrc.org/permissions')
 
+        # If the token is a service token and has the right scope, we will
+        # auth it as equivelant to an admin user
+        if ('gty' in token
+                and token['gty'] == 'client-credentials'
+                and token['scope'] == settings.CLIENT_ADMIN_SCOPE):
+            roles = ['ADMIN']
+            groups = []
+
         # Don't allow if it looks like the token doesn't have correct fields
         if groups is None or roles is None:
             return AnonymousUser()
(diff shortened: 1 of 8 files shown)


creator/settings/development.py
```diff
@@ -173,3 +173,5 @@
 AUTH0_AUD = 'https://kf-study-creator.kidsfirstdrc.org'
 CACHE_AUTH0_KEY = 'AUTH0_PUBLIC_KEY'
 CACHE_AUTH0_TIMEOUT = 86400
+
+CLIENT_ADMIN_SCOPE = 'download:file'
```
znatty22: I thought we wanted this to represent more of a role? Like `role:admin`. If not, then my next questions would be: should this be called `SERVICE_CLIENT_SCOPE` and should it be set to `'download:file upload:file'`

dankolbman: We can set it to whatever we want to name it, `role:admin` may be more representative.

dankolbman: Renamed

creator/middleware.py
```diff
@@ -169,6 +169,16 @@ def get_jwt_user(request):
         # Currently unused
         permissions = token.get('https://kidsfirstdrc.org/permissions')
 
+        # If the token is a service token and has the right scope, we will
+        # auth it as equivelant to an admin user
+      if (token.get('gty') == 'client-credentials' and
+                settings.CLIENT_ADMIN_SCOPE in token.get('scope').split())
+          )
+                and token['gty'] == 'client-credentials'
```
znatty22: These two lines snuck in, 🔥

PR #136, opened 2019-03-27
✨ Add deleteFile mutation

### LINKED ISSUE #123: Allow ADMIN to delete files
It would be helpful if the ADMIN could delete files. Even if this is not supported in the UI, it is useful for removing test files and cleaning up the clutter.

dankolbman:
Allows admin users to delete files.

Closes #123

### DIFF
--- a/tests/test_mutation.py
+++ b/tests/test_mutation.py
@@ -1,5 +1,5 @@
 import pytest
-from creator.files.models import Object, DownloadToken
+from creator.files.models import File, Object, DownloadToken
 
 
 update_query = """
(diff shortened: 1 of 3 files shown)


tests/test_mutation.py
```diff
@@ -189,3 +189,67 @@ def test_signed_url_mutation(
...
+    }
+    """
+    variables = {"kfId": file_id}
+    resp = api_client.post(
+        "/graphql",
+        content_type="application/json",
+        data={"query": query, "variables": variables},
+    )
+
+    print(resp.json())
```
XuTheBunny: You may want to remove the print statement

PR #138, opened 2019-03-29
✨ Add developer tokens

### LINKED ISSUE #137: Add developer tokens
Allow developers to create long lived download tokens that may be used similar to signed-url tokens, but without an expiration.
We may allow these tokens to be generated by an admin through a mutation.

dankolbman:
Adds developer tokens which are like signed-url tokens, except that they do not expire and may be used to download any file. These tokens may only be generated by admins.

Closes #137


creator/files/schema.py
```diff
@@ -257,6 +264,57 @@ def mutate(self, info, study_id, file_id, version_id=None, **kwargs):
...
+
+    def mutate(self, info, token, **kwargs):
+        """
+        Deletes a developer download token
+        """
+        user = info.context.user
+        if (user is None or
+                not user.is_authenticated or
+                'ADMIN' not in user.ego_roles):
+            raise GraphQLError('Not authenticated to mutate a file.')
```
znatty22:
```suggestion
            raise GraphQLError('Not authenticated to delete a token.')
```

creator/files/views.py
```diff
@@ -46,11 +48,18 @@ def download(request, study_id, file_id, version_id=None):
             # same file/object information, but will not consider it for now
             download_token = None
 
+        # Try to resolve a dev download token
+        try:
+            dev_token = DevDownloadToken.objects.get(token=token)
+        except DevDownloadToken.DoesNotExist:
+            dev_token = None
+
     if ((user is None                           # No user present
```
znatty22: Phew, this is confusing to me but I think it is saying:

```python
authenticated = user is not None and user.is_authenticated
user_authorized = (study_id in user.ego_groups or ADMIN) and download_token
developer_authorized = ADMIN and dev_token

if not (authenticated and (user_authorized or developer_authorized))
    return HttpResponseNotFound('Not authenticated to download the file.')
```

znatty22: This isn't necessarily a request for change, just trying to make sure I understand that logic correctly.

dankolbman: heh yes you can tell I thought that too by all the comments. Yes that logic looks correct.

PR #143, opened 2019-04-23
✨ Look for tokens in Authorization header

### LINKED ISSUE #142: pass developer tokens by request header or message body, not as a URL query argument
Right now tokens are sent via `?token=` query argument, but that's not secure and not recommended.

https://tools.ietf.org/html/rfc6750#section-5.3 says:

> Don't pass bearer tokens in page URLs:  Bearer tokens SHOULD NOT be
      passed in page URLs (for example, as query string parameters).
      Instead, bearer tokens SHOULD be passed in HTTP message headers or
      message bodies for which confidentiality measures are taken.
      Browsers, web servers, and other software may not adequately
      secure URLs in the browser history, web server logs, and other
      data structures.  If bearer tokens are passed in page URLs,
      attackers might be able to steal them from the history data, logs,
      or other unsecured locations.

dankolbman:
Both signed tokens and developer download tokens were required to be passed as url parameters. This PR will attempt to parse out those tokens from the `Authorization` header if it cannot find one in the url.
It expects that the tokens are prefixed with `Token ` followed by either the developer, or signed url token.

Closes #142


creator/middleware.py
```diff
@@ -148,7 +148,7 @@ def get_jwt_user(request):
             return user
 
         encoded = request.META.get('HTTP_AUTHORIZATION')
-        if not encoded:
+        if not encoded or 'Bearer ' not in encoded:
```
fiendish: You don't care if it's at the beginning or not?

tests/test_dev_tokens.py
```diff
@@ -47,3 +47,28 @@ def test_list_dev_tokens(
...
+    # Check that file may be downloaded with token header
+    resp = client.get(
+        f"/download/study/{study_id}/file/{file_id}",
+        HTTP_AUTHORIZATION=f"Token {token.token}",
+    )
+    assert resp.status_code == 200
+    assert resp.get("Content-Disposition") == expected_name
+    assert resp.content == b"aaa\nbbb\nccc\n"
+
+    # Check that file may not be downloaded with no token
```
fiendish: should test also that it may not be downloaded with an invalid token

dankolbman:
@fiendish you can [view by commit](https://github.com/kids-first/kf-api-study-creator/pull/143/files/a48113769d217533f40bedef64c8bfe5bd9b820b..6f4796a61a268cc75e541b6ffca2b92ba1d1d850)

PR #146, opened 2019-04-24
🐛 Use utf-8 encoding for Content-Disposition

### LINKED ISSUE #141: downloaded file name is polluted by percent codes
Spaces in file names get converted to `%20` when downloaded. File names should stay the same as the original.

Click the download links on https://kf-ui-data-tracker.kidsfirstdrc.org/study/SD_7NQ9151J/files

dankolbman:
Encodes the `filename` with utf-8 for the content disposition header.

Fixes #141

### DIFF
--- a/creator/files/views.py
+++ b/creator/files/views.py
@@ -96,7 +96,9 @@ def download(request, study_id, file_id, version_id=None):
         return HttpResponseNotFound('Problem finding the file')
 
     file_name = urllib.parse.quote(file.name)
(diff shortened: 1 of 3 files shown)


dankolbman:
@fiendish I was going by information that safari didn't support encoding, but it turned out that info was several years old.

creator/files/views.py
```diff
@@ -1,5 +1,6 @@
 import urllib
 from typing import Optional
+from django.utils.encoding import smart_text
```
fiendish: new import not used?

PR #175, opened 2019-05-23
♻️ Add version id to file name when downloaded

### LINKED ISSUE #161: Change filename header at download
The `filename` value in the response header when downloading a file should changed to:
1) Be sanitized of whitespace, replace spaces with '_'
2) Concatenate the version kf_id to the file
3) Add the version's extension automatically

This means users will get downloads with file names such as:
`MANIFEST_FV_00000000.csv`
`First_Batch.txt`

dankolbman:
This adds the version kf_id to the file name when downloading so that
the user gets a file that is named specifically for that version.
Otherwise, the user may download a file again either resulting in copies
of the same file which may be difficult to distinguish versions from.

Fixes #161

### DIFF
--- a/creator/files/views.py
+++ b/creator/files/views.py
@@ -1,3 +1,4 @@
+import os
 import urllib
 from typing import Optional
 from django.http import HttpResponse, HttpResponseNotFound, JsonResponse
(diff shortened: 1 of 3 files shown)


fiendish:
A regular will not be able to easily compare (or even understand the meaning of) KFIDs. 
I think it would be better if you used the actual upload time.

PR #182, opened 2019-05-28
✨ Add version file_name field

### LINKED ISSUE #178: Store and return version downloads with original file name
Files downloaded from the study creator should be named as they were uploaded.
To do this, we should add a `file_name` field to the `Version` model and use this name in the `attachment; filename=` header when offering it for download.

dankolbman:
This adds a `file_name` field to the `Version` model so that it may be used to store the original name of the file upon upload and later used when downloading to ensure that the downloaded file has the same name.
Changes download format to `{version.file_name}_{version.kf_id}.{ext}`

Fixes #178


dankolbman:
@fiendish I reverted back to the prefix. I felt it seemed more standard to append the version, but now I believe that most times (browsers) duplicate files have the version appended to the file name (`myfile.txt (1)`).

The reason for the migrations is because I opted to first alter the schema by adding the new field with a default value (empty string), and a second data migration to populate it. The third migration is a [squashed](https://docs.djangoproject.com/en/dev/topics/migrations/#migration-squashing) migration that will apply both in a single `migrate` command. This may not have been necessary, but I like the separation of logic for data vs schema.

PR #183, opened 2019-05-28
♻️ Remove file extension from default name

### LINKED ISSUE #181: Remove extensions from the default file names
It's not a good idea to mix up the `essence` of a file with it's contents, as they may change in later versions. `manifest.csv` may later become `manifest.txt`, but would require changing of the file's name to track that whereas simply `manifest` would have been a better name to begin with.

dankolbman:
Removes the file extension when creating a new `File` name.

Closes #181 

Merge #182 first.


dankolbman:
Will close in favor of #184

PR #227, opened 2019-08-06
✨ Create CavaticaProject model

### LINKED ISSUE #216: Create CavaticaProject model
We need to add a django model for a new Cavatica Project table.

CavaticaProject

XuTheBunny:
Closes #216


creator/projects/schema.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Updates an existing Cavatica project.
+        User must be authenticated and be ADMIN.
+        """
+        user = info.context.user
+        if user is None or not user.is_authenticated:
+            raise GraphQLError("Not authenticated to mutate a Cavatica project.")
+
+        try:
+            cavaticaProject = CavaticaProject.objects.get(project_id=project_id)
```
dankolbman: Maybe just change this to camel case `cavatica_project`.

creator/projects/schema.py
```diff
@@ -0,0 +1,164 @@
...
+        except CavaticaProject.DoesNotExist:
+            raise GraphQLError("Cavatica project does not exist.")
+
+        if "ADMIN" not in user.ego_roles:
```
dankolbman: Probably better to do the auth check as soon as possible and before looking up the object in the database.

creator/projects/schema.py
dankolbman: I think you need to get the `project_id` from the `kwargs` here, too.

creator/projects/schema.py
```diff
@@ -0,0 +1,164 @@
...
+            cavaticaProject.save()
+        except ClientError:
+            raise GraphQLError("Failed to save create Cavatica project")
+
+        return CreateCavaticaProjectMutation(success=True, cavaticaProject=cavaticaProject)
+
+
+class CavaticaProjectQuery(object):
+    cavaticaProject = relay.Node.Field(CavaticaProjectNode)
+    cavatica_project_by_project_id = Field(CavaticaProjectNode, project_id=String(required=True))
```
dankolbman: I think the proper 'graphQL' way to do this is to use a filter on the `allProjects` query. We can look into this.

PR #312, opened 2019-10-24
🐛 Fix the user email/slack notify toggle mutation

XuTheBunny:
Bug:
Can only set email/slack notify to `True` but cannot set it to `False`

Reason:
```
if kwargs.get("slack_notify"): <--- Here is FALSE when try to set slack_notify to FALSE
    user.slack_notify = kwargs.get("slack_notify") <--- Then this line is not excuted
```

Fix:
```
if "slack_notify" in kwargs: <--- Check if kwargs has key "slack_notify"
    user.slack_notify = kwargs.get("slack_notify")
```

### DIFF
--- a/tests/users/test_profile.py
+++ b/tests/users/test_profile.py
@@ -86,7 +86,7 @@ def test_hidden_fields(db, admin_client, field):
     "user_type,expected",
     [("admin", True), ("service", False), ("user", True), (None, False)],
 )
-def test_my_profile_mutation(
+def test_my_profile_mutation_add_notification(
     db, admin_client, service_client, user_client, client, user_type, expected
 ):
     """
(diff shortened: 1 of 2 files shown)


tests/users/test_profile.py
```diff
@@ -122,9 +122,40 @@ def test_my_profile_mutation(
...
         assert User.objects.first().slack_notify
+        assert User.objects.first().email_notify
         assert User.objects.first().slack_member_id == "U123"
     else:
         assert (
             resp.json()["errors"][0]["message"]
             == "Not authenticated to mutate profile"
         )
+
+    resp_false = api_client.post(
```
dankolbman: You could move this to a new test.

PR #328, opened 2020-01-27
✨ Allow users to specify and name research projects on creation

XuTheBunny:
- Change the field type for workflowType from enum to string
- Add projectType to be a input field for creating project
- Have special character checking and raise graphql error if invalid
Screen Shot 2020-01-29 at 2 11 20 PM

From data-tracker on creating a project: PR: https://github.com/kids-first/kf-ui-data-tracker/pull/570
- Allow user to choose one of the project type (Analysis, Delivery, Research)
- If chose Research, allow user to choose or input their customized workflow type from the workflow dropdown
- Provide special character checking for new workflow type input (no special characters allowed, only letters, numbers, space and "-")


creator/projects/schema.py
```diff
@@ -132,7 +124,16 @@ def mutate(self, info, input):
             raise GraphQLError(
                 f"Study already has a {input['workflow_type']} project."
             )
-        project = create_project(study, "HAR", input["workflow_type"])
+
+        regex = re.compile("[~`!@#$%^&*()+}{:;<>?/\\\\|]")
```
dankolbman: I think we may need to also blacklist `_` because it's not valid for urls.

XuTheBunny: Updated

PR #424, opened 2020-06-23
🔥 Remove default Cavatica projects for new studies

XuTheBunny:
Screen Shot 2020-06-26 at 6 26 20 PM

For not creating any default Cavatica projects for new studies, set the env to be:
`CAVATICA_DEFAULT_WORKFLOWS=""`

Adding default Cavatica projects for new studies, set the env to be:
`CAVATICA_DEFAULT_WORKFLOWS="workflow1 workflow2"`

Ref https://github.com/kids-first/kf-ui-data-tracker/issues/768

### DIFF
--- a/tests/projects/test_cavatica.py
+++ b/tests/projects/test_cavatica.py
@@ -29,7 +29,18 @@ def mock_create_project(mocker):
     return create_project
 
 
-def test_correct_projects(db, mock_create_project):
+def test_correct_projects_no_default(db, mock_create_project):
+    study = Study(kf_id="SD_00000000", name="test")
+    study.save()
+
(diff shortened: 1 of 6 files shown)


tests/projects/test_cavatica.py
```diff
@@ -35,13 +35,9 @@ def test_correct_projects(db, mock_create_project):
 
     setup_cavatica(study)
 
-    assert mock_create_project.call_count == 3
+    assert mock_create_project.call_count == 1
```
dankolbman, line 38: Could you add another test just like this that overrides the default projects and asserts that they are created?

PR #438, opened 2020-07-30
✨ Specify what change was made to a document in event message

### LINKED ISSUE #435: Specify what change was made to a document
When a document is changed, we should specify if the name, description, or tags were updated in the event.

https://github.com/kids-first/kf-api-study-creator/blob/4266f2ff97c984a944253bb6db24786081016ae6/creator/files/schema/file.py#L198

### LINKED ISSUE #436: Strings should be f-strings
The number of studies should be f-strings in these messages.

https://github.com/kids-first/kf-api-study-creator/blob/4266f2ff97c984a944253bb6db24786081016ae6/creator/referral_tokens/schema.py#L186-L191

XuTheBunny:
image
Closes #435

### DIFF
--- a/creator/files/schema/file.py
+++ b/creator/files/schema/file.py
@@ -181,21 +181,34 @@ class FileMutation(graphene.Mutation):
         ):
             raise GraphQLError("Not allowed")
 
+        update_fields = []
+
         try:
             if kwargs.get("name"):
+                if file.name != kwargs.get("name"):
+                    update_fields.append("name")
                 file.name = kwargs.get("name")
             if kwargs.get("description"):
+                if file.description != kwargs.get("description"):
+                    update_fields.append("description")
                 file.description = kwargs.get("description")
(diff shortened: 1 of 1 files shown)


creator/files/schema/file.py
```diff
@@ -181,21 +181,34 @@ def mutate(self, info, kf_id, **kwargs):
...
+                if file.tags != kwargs.get("tags"):
+                    update_fields.append("tags")
                 file.tags = kwargs.get("tags")
             file.save()
         except ClientError:
             raise GraphQLError("Failed to save file mutation.")
 
         # Make an update event
-        message = f"{user.username} updated file {file.kf_id}"
+        message = (
```
dankolbman, line 208: I suppose this is ok to leave an event even though none of the mutable fields are modified. The updated field is still modified so I would say it's ok, though we may consider not updating at all or making an event if nothing changed.

PR #472, opened 2020-09-08
✨ Add download extract config endpoint

### LINKED ISSUE #465: Add download extract config endpoint
We should provide an endpoint that generates an extract configuration file for a given version or document. This endpoint will be similar to the existing download endpoints and should support either document or document and version paths: https://github.com/kids-first/kf-api-study-creator/blob/master/creator/files/views.py

The views should be stored in a new `extract_configs` application under the root `creator` folder.

The endpoint should return a `.py` extract config file that has been generated from a template using Django's [template language](https://docs.djangoproject.com/en/3.1/ref/templates/language/). The template used will be based on the version's document's `file_type`. 
If the document's `file_type` has no template, no contents should be returned with a 404 status.

XuTheBunny:
image

Closes #465


creator/extract_configs/views.py
```diff
@@ -0,0 +1,114 @@
...
+        return HttpResponse(
+            "Not authorized to extract config for the file", status=401
+        )
+
+    # Don't return anything if the file does not belong to the requested study
+    if file.study.kf_id != study_id:
+        return HttpResponseNotFound("No file exists for given ID and study")
+
+    # Need to set storage location for study bucket if using S3 backend
+    if settings.DEFAULT_FILE_STORAGE == "django_s3_storage.storage.S3Storage":
```
dankolbman, line 101: We don't need this because we don't need to ever download the contents of the version in this endpoint.

creator/extract_configs/views.py
```diff
@@ -0,0 +1,114 @@
...
+        obj.key.storage = S3Storage(aws_s3_bucket_name=file.study.bucket)
+
+    # If we're using a token, mark it as claimed
+    if download_token:
```
dankolbman, line 105: I need to revisit how download tokens work. They may not be valid for things other than versions or documents.

creator/extract_configs/views.py
```diff
@@ -0,0 +1,116 @@
...
+        download_token is None and dev_token is None
+    ):
+        return HttpResponse(
+            "Not authorized to extract config for the file", status=401
+        )
+
+    # Don't return anything if the file does not belong to the requested study
+    if file.study.kf_id != study_id:
+        return HttpResponseNotFound("No file exists for given ID and study")
+
```
dankolbman, line 99: Before we attempt to download the file, we need to make sure that the file has an extract config template. If it does not, we should return a 404.

creator/extract_configs/views.py
```diff
@@ -0,0 +1,116 @@
...
+    # If we're using a token, mark it as claimed
+    if download_token:
+        download_token.claimed = True
+        download_token.save()
+
+    content = render_to_string(
+        "s3_scrape_config.py",
```
dankolbman, line 106: We need to be able to change what template is used based on the file type. We could add another field to the FILE_TYPES to specify what template to use. If the file type does not have a template, we can just use `None` and return 404 for the extract config requests.
https://github.com/kids-first/kf-api-study-creator/blob/master/creator/analyses/file_types.py

creator/extract_configs/views.py
```diff
@@ -0,0 +1,116 @@
...
+        {"download_url": f"https://{request.get_host()}{obj.path}"},
```
dankolbman, line 107: It may be more general if you instead pass in the file and or version here. Then you can use `{{ version.download_url }}` or access any other properties on the file or version within the template.

creator/extract_configs/views.py
```diff
@@ -0,0 +1,122 @@
...
+        obj = file.versions.latest("created_at")
+    return file, obj
+
+
+def download_config(request, study_id, file_id, version_id=None):
+    """
+    Allow download if user is an admin OR user belongs to the file's study
+    OR if there is an unclaimed token provided that has not expired.
+    """
+    TYPE_TEMPLATE = {"S3S": "s3_scrape_config.py"}
```
dankolbman, line 33: Can you put this into the FILE_TYPES in here? https://github.com/kids-first/kf-api-study-creator/blob/master/creator/analyses/file_types.py
Though we may eventually want to move that file to the settings.

creator/extract_configs/views.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Don't return anything if the file does not belong to the requested study
+    if file.study.kf_id != study_id:
+        return HttpResponseNotFound("No file exists for given ID and study")
+
+    # Check if the file type is valid for extracting config
+    if FILE_TYPES[file.file_type]["template"] is not None:
+        template = FILE_TYPES[file.file_type]["template"]
+    else:
+        return HttpResponseNotFound("Not valid for extracting config")
```
dankolbman, line 105: Can you update this to `Cannot generate extract config for the {file.file_type} file type`

dankolbman (approved):
:rabbit:

PR #482, opened 2020-09-23
✨ Job logs

### LINKED ISSUE #476: Save task logs to s3
The logs created by our task runners that consume the redis queue don't store their logs anywhere and are only visible by digging through cloudwatch logs.
We should send task logs to the disk and upload them upon task completion instead.

### LINKED ISSUE #477: Retrieve task logs from s3
We need to be able to retrieve logs from s3 for task runs so that we can present them to the user for evaluation.
This should be done in the form that we use for file and version downloads.

dankolbman:
Saves logs for jobs.

Closes #476 
Closes #477


XuTheBunny (approved):
🥩

PR #483, opened 2020-09-24
🐛 Stringify before checking for printable chars

dankolbman:
If there are column names that are non string, like numbers, the current method will fail to parse because `isprintable()` is a string method.


XuTheBunny (approved):
🥨

PR #486, opened 2020-10-08
✨ Releases application

### LINKED ISSUE #487: Implement release resources
Integrating the Release Coordinator functionality into the Study Creator will require the essential models to be implemented.
These models are:
 - Releases
 - Tasks
 - Task Services

These will need to contain the same fields as in the Coordinator, at a minimum, so as to be backward compatible.

### LINKED ISSUE #488: Sync Release Coordinator data periodically
While the role of releasing data slowly moves to the Study Creator API, new releases created in the Release Coordinator will need to be synchronized. This will offer a window where releases may be created in either the Coordinator or the Study Creator.
This task to sync releases will eventually be phased out and removed once all new releases are created in the Study Creator.

dankolbman:
Adds a releases application to implement Release Coordinator functionality in the Study Creator API.

- Adds tracebacks to job logs
- Add releases application with concepts:
    - Release
    - Release Task
    - Release Service 
- Add release synchronization to back-port data from Release Coordintaor periodically
- Add documentation for releases
- Add release events
- Add slack notifications

Closes #487
Closes #488


creator/releases/models.py
```diff
@@ -0,0 +1,429 @@
...
+        related_name="tasks",
+    )
+    release_service = models.ForeignKey(
+        "ReleaseService",
+        on_delete=models.CASCADE,
+        null=False,
+        blank=False,
+        related_name="tasks",
+    )
+    job_log = models.ForeignKey(
```
znatty22, line 202: This relation seems a little odd to me because its indirectly linking two things that maybe should be linked directly. I think of it like: 

- A Task is a job definition
- A Task is run to produce a Job
- A Job has a JobLog

So with this in mind, I would think you'd have a Task relate to a Job which relates to a JobLog?

creator/releases/models.py
```diff
@@ -0,0 +1,492 @@
...
+        except requests.exceptions.RequestException as err:
+            logger.error(f"problem requesting task for {action}: {err}")
+            raise err
+
+        try:
+            state = resp.json()
+            logger.info(
+                f"Recieved response from {self.release_service.url}: {state}"
+            )
+        except ValueError as err:
```
znatty22, line 255: Shouldn't this be `json.decoder.JSONDecodeError`?

creator/releases/mutations.py
```diff
@@ -0,0 +1,329 @@
...
+        release.save()
+        release.studies.set(studies)
+
+        # Now make tasks for each service since we know they all exist
+        # and the release has been created to tie them to
+        for service in services:
+            task = ReleaseTask(release=release, release_service=service)
+            task.save()
+
+        release.initialize()
```
znatty22, line 95: In all of these mutations, should the state change + save happen inside the rq job?

creator/releases/tasks.py
```diff
@@ -0,0 +1,574 @@
...
+            f"The release will be automatically moved to the '{target}' state."
+        )
+        if failed:
+            release.failed()
+        else:
+            release.canceled()
+        release.save()
+        return
+
+    # Iterate through each task sequentially and tell it to publish
```
znatty22, line 410:
```suggestion
    # Iterate through each task sequentially and tell it to cancel
```

creator/releases/tasks.py
```diff
@@ -0,0 +1,574 @@
...
+
+        queue = django_rq.get_queue("releases")
+        queue.enqueue(
+            cancel_release,
+            release_id=task.release.pk,
+            failed=True,
+            ttl=settings.RQ_DEFAULT_TTL,
+        )
+
+    # Check if all tasks are initialized now and queue up the release start
```
znatty22, line 243: I'm wondering why this block is here instead of at the end of the `initialize_release` task? It looks like the `start_release` task will be queued for each task when it should be queued once for just the release?

znatty22: Still wondering :)

znatty22: Ok I see what you're doing now. My second thought was could `initialize_task` just be a method that gets called in the task loop on line 201, but that would make the initializing all of the tasks a synchronous process which I presume you don't want

creator/decorators.py
```diff
@@ -26,9 +28,14 @@ class task:
     """
     A decorator to uniformly setup tasks that get enqueued for workers to
     execute.
+
+    If the wrapped task contains a release_id or task_id, save the log to the
+    appropriate model so that it may be retrieved relationally.
     """
 
     def __init__(self, job=None):
+        self._release = None
```
znatty22, line 37: I think you may be able to generalize the task decorator a bit. It looks like this method does the same thing for both Release and ReleaseTask (and later will also include IngestRun) - attach a JobLog to the object. 

So rather than passing in the foreign key parameter for each model class, could you have one keyword arg called `attach_log_to` (or something else?) which would take in a list of tuples: `[(Release, <the release id>), (ReleaseTask, <the release task id>)]` and then attach JobLogs to each model class instance if it exists (so basically what you do now for release and task)

znatty22 (changes requested):

creator/decorators.py
```diff
@@ -29,11 +29,22 @@ class task:
...
     execute.
 
-    If the wrapped task contains a release_id or task_id, save the log to the
-    appropriate model so that it may be retrieved relationally.
+    related_models may be specified to attach the generated job_log to other
+    models.
+    This takes one of two forms:
+        {Model1: "kwarg_name_for_pk", Model2: ...}
+            or
+        {Model1: ModelInstance, Model2: ...}
```
znatty22, line 37: What's the use case for this?

creator/decorators.py
```diff
@@ -29,11 +29,22 @@ class task:
...
+    table in order to be saved.
+    If there is no foreign key, or no object instance can be resolved, it will
+    be skipped and no relation saved.
+    When calling the job function, it is important that the appropriate keyword
+    argument is passed, positional arguments will not be able to be resolved.
     """
 
-    def __init__(self, job=None):
+    def __init__(self, job=None, related_models=None):
         self._release = None
```
znatty22, line 48: Probably want to rm self._release, self._task now

creator/decorators.py
```diff
@@ -167,13 +216,7 @@ def close(self):
...
-        # Save the log to the release and/or task, if there is one
-        if self._release:
-            self._release.job_log = log
-            self._release.save(update_fields=["job_log"])
-        if self._task:
-            self._task.job_log = log
-            self._task.save(update_fields=["job_log"])
+        self._attach_related(log)
 
     def _get_existing_job_log(self):
```
znatty22, line 221: I still see references to self._release, self._task in this method

znatty22 (changes requested):

creator/decorators.py
dankolbman: So we can pass model instances in directly or by primary key:
```
release = Release()
my_job(release=release)
my_job(release=release.pk)
```
Both of these are valid.

creator/decorators.py
```diff
@@ -182,11 +223,12 @@ def _get_existing_job_log(self):
...
 
-        if self._release and self._release.job_log is not None:
-            return self._release.job_log
-
-        if self._task and self._task.job_log is not None:
-            return self._task.job_log
+        # Grab the first job log we find and return it
+        # This could have unforseen implications if the related instances have
+        # different job logs: only the first's log contents will be appended to
+        for instance in self.related_instances:
```
znatty22, line 229: 🤔 Don't you want to update all related instances' job logs, not just the first one you find?

dankolbman: yeah that's probably the way to go.

creator/decorators.py
znatty22: Oh oh I see. I kept imagining `task({Release: '182793628136'})` but that makes no sense

znatty22 (approved):
🦃 ✅

PR #494, opened 2020-10-22
✨ Add Genomic Workflow Output Manifest

### LINKED ISSUE #492: New expedited document type for genomic workflow output data
We need to add a new expedited document type to capture the genomic workflow output manifest produced by the Bix team after they run the genomic workflows on the source sequencing files within a Study. This manifest will be used by the ingest job to register the harmonized genomic files and link them to their specimens and sequencing experiments in the Dataservice.

The current manifest format is a JSON file that looks like:  https://datatracker.kidsfirstdrc.org/study/SD_8Y99QZJJ/documents/SF_KQZ9MMH6

We would like to change this format to a tabular format so that no additional changes are needed on Data Tracker.

**To Do**

- [x] Verify with Bix that the name "Genomic Workflow Output Manifest" accurately represents the concept
- [x] Design the desired tabular format for the Genomic Workflow Output Manifest and get it approved by Bix
- [x] Add functionality to Study Creator API for new expedited type: Genomic Workflow Output Manifest

znatty22:
Closes #492


dankolbman:
Can you name your migration with the `-n` flag?

creator/analyses/file_types.py
```diff
@@ -50,4 +50,19 @@
...
+            'cavatica_task',
+            'kf_biospecimen_id',
+            'kf_participate_id',
+            'kf_family_id',
+            'filepath',
+            'data_type',
+            'workflow_type',
+            'source_read'
+        ],
+        "template": "genomic_workflow_output_config.py",
```
dankolbman, line 66: I think we need to make this template or else we may hit bugs when we try to return an extract config for this file type.

znatty22: Ok will add

dankolbman (approved):
# 🕊

PR #495, opened 2020-10-24
🐛 Fix startresource command

znatty22:
I needed to make a couple changes in the templates and the template generation code in order to get the tests to pass.

1. Use lower/upper camel casing in test query/mutation templates
    - Can't use Python's str.title() method because it doesn't turn snake case into proper camel case (e.g. ingest_request -> Ingest_Request)
2. Change permission codenames referenced in the code to match what gets generated in the db
    - When django creates permissions, it uses the lower cased version of the model name (e.g. list_all_ingestrequest)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
           name: Run code tests
           working_directory: ~/workdir
           command: |
-            pytest
+            pytest tests
             codecov --token=$CODECOV_TOKEN
       - run:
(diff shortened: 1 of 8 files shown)


dankolbman (approved):
👍 
Thanks for these changes. I'm not sure if we'll ever merge this all back to master, maybe if it proves useful enough.

znatty22:
Thanks for making this, I thought was very useful!

PR #497, opened 2020-11-03
✨ Add initial ingest run resource

### LINKED ISSUE #499: Add an IngestRun model
We need a new Django model (IngestRequest) in the Ingest Requests app (see #491) to capture a request that the user makes to ingest file(s) into the Kids First Data Service. 

An IngestRequest will have a many to many relationship with the Version model so that a single request can be used to ingest one or more files. There should only be 1 IngestRequest per unique set of file Versions since there is no sense in ingesting the same set of content if that content is already being ingested. Last, an IngestRequest will have a 1 to 1 relationship with a Job to track the execution of the associated ingest task.

znatty22:
Closes #499


creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
-    Plug the model in here
+    Request to ingest file(s) into a target data service
     """
 
     class Meta:
         permissions = [
             ("list_all_ingestrequest", "Show all ingest_requests"),
         ]
 
+    name = models.CharField(
```
dankolbman, line 25: Is this being used as a primary key, or a way to constrain uniqueness so that you may only have one ingest run for a particular set of versions?

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+        on_delete=models.SET_NULL,
+        help_text="The user who submitted this ingest request",
     )
+
+    @staticmethod
+    def name_from_version_ids(version_ids):
+        """
+        Create IngestRequest.name by joining the version ids together
+        """
+        return DELIMITER.join([NAME_PREFIX] + version_ids)
```
dankolbman, line 71: If you're trying to create a unique key for this ingest request that's reproducible, you probably want to sort your ids first.
I would suggest hashing the concatenated string instead, though, cause I assume there will be no limit on the number of versions in an ingest run.

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+    @staticmethod
+    def version_ids_from_name(request_name):
+        """
+        Parse IngestRequest.name into the file version ids which were used
+        to create the name
+        """
+        return request_name.lstrip(NAME_PREFIX).split(DELIMITER)
+
+    @classmethod
+    def find_by_version_ids(cls, version_ids):
```
dankolbman, line 82: What are you using this method for?
I think it's a lot less obscure if you don't abstract the one liner query into this static method.

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+            [v.kf_id for v in instance.versions.all()]
+        )
+
+    def _versions_exist(self):
+        """
+        Check that an IngestRequest has at least 1 Version associated with it.
+        """
+        # If instance hasn't been saved yet, then there is no possibility
+        # for versions to exist
+        if not self.id:
```
dankolbman, line 134: I think there may be a more appropriate way to check if it's still waiting to be written. Maybe `._dirty` or something similar.

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+            return False
+
+        # If no versions have been added, instance.versions won't exist
+        versions_exist = True
+        try:
+            self.versions
```
dankolbman, line 140: I believe the more semantic way of doing this would be to check `self.versions.exists()` rather than throwing an exception.
Have you tested this? I would expect `self.versions` to return a queryset that resolves to an empty list rather than throw an exception.

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+            raise ValidationError(
+                {"versions": "An IngestRequest must have at least 1 Version"}
+            )
+
+    @staticmethod
+    def _pre_save_hook(sender, instance, *args, **kwargs):
+        """
+        Set the IngestRequest.name using the version ids
+        """
+        # If no versions have been added yet, do nothing
```
dankolbman, line 119: Can you simply run the `clean` function here instead?

creator/ingest_requests/models.py
```diff
@@ -1,16 +1,160 @@
...
+        This method is called when an IngestRequest is deleted
+        """
+        pass
+
+    def clean(self):
+        """
+        Validate model instance and raise a
+        django.core.exceptions.ValidationError if validation fails
+        """
+        if not self._versions_exist():
```
dankolbman, line 109: It seems like this could more easily be replaced by `if not self.versions.exists():`

creator/ingest_requests/models.py
dankolbman: Constructing this field seems to create some complications in creating the model. Could it instead be a computed property?

```
@property
def name(self):
    return ''.join(v.name for v in self.versions.all())
```

creator/ingest_requests/mutations.py
```diff
@@ -37,12 +43,49 @@ def mutate(self, info, input):
...
+                # Add versions to ingest request
+                ingest_request.versions.set(versions)
+                # Validate ingest request
+                ingest_request.full_clean()
+                ingest_request.save()
+        except ValidationError as e:
+            raise GraphQLError(f"Ingest request validation error: {str(e)}")
+        except IntegrityError as e:
+            if "duplicate key value" in str(e):
+                raise GraphQLError(
```
dankolbman, line 71: Why can you only run an ingest once for a unique set of versions?
If my first ingest fails, shouldn't I be able to retry it?
Or if some later ingest overwrites this one, shouldn't I be able to run it again?

creator/ingest_requests/models.py
znatty22: Hmm I must have been doing something wrong before bc I was getting this exception, a lot. You are right tho, I just did a quick test and I can just get the queryset and check the length. I wonder if its bc I was playing around in the django shell and didn't have autoreload on.

creator/ingest_requests/mutations.py
znatty22: I think it depends on how you think of an IngestRequest. I went back and forth on the concept of it but right now I'm thinking of itas a set of input parameters for an ingest job. Currently those parameters just consist of the set of files that will be ingested in the job. You should be able to execute an ingest job for a set of input parameters as 
many times as you want, but I don't think it makes sense to have multiple ingest jobs running at the same time for the same set of input parameters. I think that would be equivalent to running `kidsfirst ingest /path/to/ingest_package` on your machine in two different processes.

creator/ingest_requests/models.py
znatty22: Yea I considered hashing it but I ultimately decided to keep it simple. I could consider that again though if we really want to keep name to a fixed length and store it more efficiently as a shorter string. 

True, there is no limit to the files that can be part of an ingest request (technically, we should be saying an ingest package) but realistically we will probably never ingest more than ~50 files in an ingest request which is why I just restricted name to a max length of 500 (11 char version id * ~50 files).

creator/ingest_requests/models.py
znatty22: It's not used yet but it will be used in the create ingest mutation in a future PR (see #500)

tests/ingest_runs/test_model.py
```diff
@@ -0,0 +1,35 @@
...
+        assert not ir.input_hash
+        assert not ir.name
+        ir.versions.set(file_versions)
+        ir.save()
+        assert ir.input_hash
+        assert ir.name
+        for v in file_versions:
+            assert v.kf_id in ir.name
+    ingest_runs = [(ir.input_hash, ir.name) for ir in IngestRun.objects.all()]
+    ingest_runs[0] == ingest_runs[1]
```
gsantia, line 27: should this be
`assert ingest_runs[0] == ingest_runs[1]` ?

creator/ingest_runs/models.py
```diff
@@ -1,16 +1,110 @@
...
+            "Identifies an ingest run by its input parameters. Autopopulated "
+            "on save with the MD5 hash of all of the ingest run "
+            "input parameters."
+        ),
+    )
+    name = models.TextField(
+        blank=True,
+        null=True,
+        help_text=(
+            "The name of the ingest run. An field that is auto populated "
```
gsantia, line 42:
```"The name of the ingest run. A field that is auto-populated "

File: tests/ingest_runs/test_model.py (hunk 9 above)
Comment (znatty22) [reply tests/ingest_runs/test_model.py:27]: Haha oops

File: tests/ingest_runs/test_ingest_runs_mutations.py [hunk 11]
```diff
@@ -37,31 +61,51 @@
...
-    )
+    # Create a study with some files
+    for i in range(2):
+        prep_file(authed=True)
+    version_ids = [
+        to_global_id("VersionNode", f.versions.first().kf_id)
+        for f in File.objects.all()
+    ]
+
+    # Start a valid ingest runs for a batch of file versions
```
gsantia, line 80: `# Start valid ingest runs for a batch of file versions`

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -37,31 +61,51 @@
...
-        assert (
-            resp.json()["data"]["createIngestRun"][
-                "ingestRun"]
-            is not None
-        )
+        ir = resp.json()["data"]["startIngestRun"]["ingestRun"]
+        assert ir["name"]
+        assert ir["inputHash"]
+        for version in ir["versions"]["edges"]:
+            version["node"]["kfId"] in ir["name"]
```
gsantia, line 88: `assert version["node"]["kfId"] in ir["name"]

tests/ingest_runs/test_ingest_runs_queries.py
```diff
@@ -63,18 +105,63 @@ def test_query_ingest_run(db, clients, user_group, allowed):
         (None, False),
     ],
 )
-def test_query_all_ingest_runs(db, clients, user_group, allowed):
+def test_query_all_ingest_runs(db, clients, ingest_runs, user_group, allowed):
     client = clients.get(user_group)
 
-    ingest_run = IngestRunFactory.create_batch(5)
+    # Create an ingest run for each file
+    n = 2
```
gsantia, line 112: ?

creator/ingest_runs/migrations/0001_initial.py
```diff
@@ -1,24 +1,95 @@
...
+                        editable=False,
+                        help_text="Identifies an ingest run by its input parameters. Autopopulated on save with the MD5 hash of all of the ingest run input parameters.",
+                        null=True,
+                    ),
+                ),
+                (
+                    "name",
+                    models.TextField(
+                        blank=True,
+                        help_text="The name of the ingest run. An field that is auto populated by concatenating the IDs of the IngestRun's file versions",
```
gsantia, line 45: `help_text="The name of the ingest run. A field that is autopopulated by concatenating the IDs of the IngestRun's file versions",`

creator/ingest_runs/models.py
```diff
@@ -39,8 +39,8 @@ class Meta:
         blank=True,
         null=True,
         help_text=(
-            "The name of the ingest run. An field that is auto populated "
-            "by concatenating the IDs of the IngestRun's file versions"
+            "The name of the ingest run. Auto populated on save with the "
```
gsantia, line 42: `The name of the ingest run. Autopopulated on save with the`

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -77,15 +77,15 @@ def test_start_ingest_run(
...
+    # Start valid ingest runs for a batch of file versions
     resp = send_query(client, START_INGEST_RUN, {"versions": version_ids})
 
     if allowed:
         ir = resp.json()["data"]["startIngestRun"]["ingestRun"]
         assert ir["name"]
         assert ir["inputHash"]
         for version in ir["versions"]["edges"]:
-            version["node"]["kfId"] in ir["name"]
+           assert version["node"]["kfId"] in ir["name"]
```
gsantia, line 88: missing a space here

znatty22: Oops

gsantia (approved):

PR #498, opened 2020-11-03
✨ New ingest runs app

### LINKED ISSUE #491: Ingest genomic workflow output data into Dataservice
Analysts currently spend too much manual effort and time to ingest the genomic workflow outputs into the Dataservice. 
This epic is about adding functionality to the Study Creator API to ingest the genomic workflow output data.  Additionally, UI elements will be added to Data Tracker so that Analysts may push a button to trigger the Study Creator ingestion.

Ingestion of workflow output data includes: 
- Registration of workflow output genomic files in Dataservice
- Linking of genomic files to their Biospecimens in Dataservice
- Linking of genomic files to their Sequencing Experiments in Dataservice

### LINKED ISSUE #493: Initial Django app to ingest genomic workflow output data
We will need a new Django application within the Study Creator API for ingest related functionality. To start, this application will include the necessary functionality for ingesting the data in the Genomic Workflow Output Manifest in #492. Later on we may add more functionality such as data validation and ingestion of other documents.

### LINKED ISSUE #499: Add an IngestRun model

### LINKED ISSUE #500: Add ingest run jobs

Implement:

### Ingest Run Life Cycle 

enqueues a run ingest task.

### Events

### Logging

### LINKED ISSUE #550: Validation Models
## Overview

Related to #548 

A DataReview will involve one or more validation runs on the files in the data review. We will need to capture the concept of a validation run somehow. This might mean modifying/extending the IngestRun model if we treat a validation run as a kind of IngestRun or it might mean creating an entirely new model.

It is very likely that we will represent a validation run as an ingest run with a specific set of input parameters that result in an ingest run where it terminates after the validation step.

## Tasks
-   [x] Create models to capture the concept of a validation run and validation results

## Design Considerations

- A Data Review may have one or more validation runs for the files in the review
- A Validation Run will produce a set of validation results

znatty22:
Closes #493
Closes #499 
Closes #500
Closes #550 
Closes #605

This PR introduces the new Ingest Runs app which will encompass functionality to automate various ingest processes such as ingesting harmonized genomic files (#491). For now this PR only sets up the app with placeholder ingest tasks that will be filled in once #501 is complete.


dankolbman (changes requested):
It would be good to get those last couple of missing tests incorporated and maybe add in the feature flags to conditionally do ingests.
See here for missing tests: https://app.codecov.io/gh/kids-first/kf-api-study-creator/compare/498/diff

creator/ingest_runs/signals.py
```diff
@@ -0,0 +1,84 @@
...
+    cancel_invalid_ingest_runs(instance)
+
+
+def cancel_invalid_ingest_runs(version):
+    """
+    Cancel all currently running IngestRuns that contain _version_ among its
+    Versions.
+    """
+    invalid_ingests = IngestRun.objects.filter(
+        state="started", versions=version
```
znatty22, line 45: @gsantia As I'm implementing validation runs I noticed this bug. This line will find ingest runs whose versions equal the input version, but what we really want is ingest runs whose versions contain the input version. I think this line needs to change to:
```suggestion
        state="started", versions__in=version
```

See https://docs.djangoproject.com/en/3.1/ref/models/querysets/#in

znatty22, line 45: @gsantia As I'm implementing validation runs I noticed this bug. This line will find ingest runs whose versions equal the input version, but what we really want is ingest runs whose versions contain the input version. I think this line needs to change to:
```suggestion
        state="started", versions__contains=version
```

See https://docs.djangoproject.com/en/3.1/ref/models/querysets/#contains

znatty22, line 45: @gsantia As I'm implementing validation runs I noticed that this line. This looks like it will find ingest runs whose versions equal the input version, not ingest runs whose versions contain the input version. Have you tested that this gets us the latter?

znatty22: Sorry never mind, apparently this works. I just tested it in django shell

dankolbman (approved):
# 💠

PR #502, opened 2020-11-04
✨ Ingest task genomic workflow output

### LINKED ISSUE #501: Add task to ingest Genomic Workflow Output Manifest
Implement a new task (`creator/tasks.py`) that will read the contents of the `Genomic Workflow Output Manifest` and ingest the necessary entities into the Kids First Data Service. This task should be implemented in an idempotent way so that it produces the same result in the Data Service no matter how many times it is run on the same input file. 

### Task Implementation

**Methods**
```
# creator/ingest_runs/tasks.py

def ingest_genomic_workflow_outputs(version_ids):
    # Implement task here
```

**Execution**
The following entities must be ingested into the Data Service: 

1. Genomic file metadata (e.g. size, hashes, etc) at Data Service endpoint `/genomic-files`
    - The file metadata for each file in the manifest will need to be scraped from S3
2. Genomic file, biospecimen links at Data Service endpoint `/biospecimen-genomic-files`
   - This information will come directly from the manifest
3. Genomic file, sequencing experiment links at Data Service endpoint `/sequencing-experiment-genomic-files`
   -  In order to figure out which sequencing experiment to link the harmonized file to, we will either have to: 
       a. Ask Bix team to add columns to the manifest to capture this information
       b. Query the Data Service in the following way:  

           ``` 
           1. Get the study's unharmonized genomic files `/genomic-files?study_id=foo&is_harmonized=False&visible=True`
           2. Get the study's sequencing experiments `/sequencing-experiments?study_id=foo&visible=True`
           3. Get the study's sequencing-experiment-genomic-file links `/sequencing-experiment-genomic-files?study_id=foo&visible=True`
           4. Get the study's biospecimen-genomic-file links `/biospecimen-genomic-files?study_id=foo&visible=True`
           5. Join the tables from 1, 2, 3, 4 together on the genomic file ID
           6. Join the merged table from step 5 with the manifest on the biospecimen ID
           
           This should produce a table that has sequencing experiment information for every harmonized genomic workflow file 
           ```

gsantia:
Closes #501 

# Todo

## First

- [x]  Natasha - Simplify study generator so it doesn't depend on S3
- [x] Gio - Clean up commit history
- [x] Gio - implement the initial ingest GWO tests with mocks (no real s3 or dataservice interactions) 

## Second
-   Detect and report missing data
    -   What happens when we merge things together and lose records?
    -   How do we catch this (use outer merge) and how do we report this
    -   Cases - Dataservice is missing specimens, seq experiments
-   Improve error handling
    -   Add try/excepts around each load function in ingest_gwo
    -   Log the exception and then re-raise to gracefully stop execution of the
        ingest gwo
-   Validate ingest_gwo 
    - We should add post ingest_gwo validation / sanity check to make sure the right # of entities were loaded
    -   Similar to \_assert_load in the study_generator.py
-   Event firing for various important things
    - Fire an event at the end of a successful ingest task that contains total counts of entities loaded
    - Fire an event when we encounter missing data
- Support genomic files on local file system or S3
    - GenomicDataLoader should be able to get file metadata for files in the GWO manifest based on protocol (file:// or s3://)
    - StudyGenerator should be able to generate file manifests for files on localhost or s3
    - Even though the tests will mock S3 methods, if a developer wants to play with the ingest GWO functionality locally (using Data Tracker), right now they cannot do it without being connected to the VPN since the ingest_gwo method will try to scrape S3.
    - If we could support ingesting a gwo manifest with local filepaths then this would eliminate the dep on S3 and developers can run everything locally.
-   Check how real Cavatica task and Cavatica project IDs are generated

### DIFF
--- a/creator/ingest_runs/genomic_data_loader.py
+++ b/creator/ingest_runs/genomic_data_loader.py
@@ -1,16 +1,15 @@
 from creator.ingest_runs import utils
+from django.conf import settings
 import ast
-import os
 import logging
-from pprint import pprint
-
+import os
 import pandas as pd
 
 from kf_lib_data_ingest.app import settings as ingest_settings
 from kf_lib_data_ingest.common.concept_schema import CONCEPT
 from kf_lib_data_ingest.etl.load.load import LoadStage
 
-from creator.ingest_runs.data_generator.ingest_package.extract_configs.s3_scrape_config import genomic_file_ext  # noqa
+from creator.studies.data_generator.ingest_package.extract_configs.s3_scrape_config import genomic_file_ext  # noqa
 
 GEN_FILE = "genomic_file"
 GEN_FILES = "genomic-files"
@@ -20,29 +19,15 @@ SEQ_EXP_GEN_FILE = "sequencing_experiment_genomic_file"
 SEQ_EXP_GEN_FILES = "sequencing-experiment-genomic-files"
 BIO_GEN_FILE = "biospecimen_genomic_file"
 BIO_GEN_FILES = "biospecimen-genomic-files"
-LOAD_ENTITY_TYPES = [GEN_FILE, BIO_GEN_FILE, SEQ_EXP_GEN_FILE]
+LOAD_ENTITY_TYPES = {GEN_FILE, BIO_GEN_FILE, SEQ_EXP_GEN_FILE}
 
-LOCALHOST = "http://localhost:5000"  # testing
-DATASERVICE_URL = LOCALHOST  # During testing
+logger = logging.getLogger(__name__)
 
 
 class GenomicDataLoader(object):
-
-    def __init__(self, study_id, init_logger=False):
-        self.study_id = study_id
-
-        if init_logger:
-            format_ = (
-                "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
-            )
-            root = logging.getLogger()
-            consoleHandler = logging.StreamHandler()
-            consoleHandler.setFormatter(logging.Formatter(format_))
-            root.setLevel("INFO")
-            root.addHandler(consoleHandler)
-
-        self.logger = logging.getLogger(type(self).__name__)
-
+    def __init__(self, study):
+        self.study = study
+        self.study_id = study.kf_id
         self.loader = None
         self.target_api_cfg = ingest_settings.load().TARGET_API_CONFIG
 
@@ -61,7 +46,7 @@ class GenomicDataLoader(object):
         source files have direct relations to the specimens and sequencing
         experiments.
         """
-        self.logger.info(
+        logger.info(
             "Begin the ingest process for genomic workflow output files"
         )
         # -- Step 1: Load harmonized genomic files --
@@ -81,7 +66,7 @@ class GenomicDataLoader(object):
     def load_harmonized_genomic_files(self, manifest_df):
         """
         Load harmonized genomic file metadata from the manifest into the
-        DataService.Scrape S3 to obtain file metadata needed.
+        DataService. Scrape S3 to obtain file metadata needed.
 
         After loading, grab the generated kf_ids from the ingest lib cache,
         join them to _genomic_df_ and then return _genomic_df_. This DataFrame
@@ -90,10 +75,8 @@ class GenomicDataLoader(object):
         assert not manifest_df.empty, "Empty manifest DataFrame"
 
         # Get S3 file metadata
-        buckets = set(
-            os.path.split(row)[0] for row in manifest_df["Filepath"]
-        )
-        self.logger.info(f"Getting file metadata from S3 {buckets} ...")
+        buckets = set(os.path.split(row)[0] for row in manifest_df["Filepath"])
+        logger.info(f"Getting file metadata from S3 {buckets} ...")
 
         file_df = utils.scrape_s3(buckets)
         file_df = file_df[~file_df["Filename"].isnull()]
@@ -106,9 +89,9 @@ class GenomicDataLoader(object):
         genomic_df = self._prep_harmonized_gf_df(genomic_df)
 
         df = genomic_df.drop_duplicates([CONCEPT.GENOMIC_FILE.ID])
-        self.logger.info(
+        logger.info(
             f"Loading {df.shape[0]} harmonized genomic files into "
-            f"{DATASERVICE_URL}"
+            f"{settings.DATASERVICE_URL}"
         )
 
         # Load harmonized genomic files
@@ -125,28 +108,32 @@ class GenomicDataLoader(object):
         the appropriate biospecimen-genomic-file records into the Data Service
         """
         # Prepare data
-        biospec_gen_df = genomic_df[[
-            CONCEPT.GENOMIC_FILE.ID,
-            CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
-            CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
-        ]]
+        biospec_gen_df = genomic_df[
+            [
+                CONCEPT.GENOMIC_FILE.ID,
+                CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
+                CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
+            ]
+        ]
         biospec_gen_df[CONCEPT.BIOSPECIMEN.ID] = biospec_gen_df[
             CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID
         ]
         biospec_gen_df[CONCEPT.BIOSPECIMEN_GENOMIC_FILE.VISIBLE] = True
         df = biospec_gen_df.drop_duplicates(
-            [CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
-             CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID],
+            [
+                CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
+                CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
+            ],
         )
 
         # Load into Data Service
-        self.logger.info(
+        logger.info(
             f"Creating {df.shape[0]} biospecimen-genomic-file links in "
-            f"{DATASERVICE_URL}"
+            f"{settings.DATASERVICE_URL}"
         )
         _ = self._load_entities(BIO_GEN_FILE, df)
 
-        return biospec_gen_df
+        return df
 
     def load_seq_exp_harmonized_genomic_files(self, genomic_df):
         """
@@ -162,11 +149,9 @@ class GenomicDataLoader(object):
         """
         # Get sequencing experiments for the unharmonized genomic files in
         # the study
-        se_source_gf_df = self._get_seq_experiment_genomic_files(
-            harmonized=False
-        )
-        # Merge ^ with the genomic df which as both harmonized and unharmonized
-        # gfs in one table
+        se_source_gf_df = self._get_seq_experiment_genomic_files()
+        # Merge ^ with the genomic df which has both harmonized and
+        # unharmonized gfs in one table
         se_gf_df = pd.merge(
             genomic_df, se_source_gf_df, on=CONCEPT.GENOMIC_FILE.SOURCE_FILE
         )
@@ -174,42 +159,45 @@ class GenomicDataLoader(object):
 
         # Load the harmonized genomic-file sequencing-experiment links
         df = se_gf_df.drop_duplicates(
-            [CONCEPT.SEQUENCING.TARGET_SERVICE_ID,
-             CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID],
+            [
+                CONCEPT.SEQUENCING.TARGET_SERVICE_ID,
+                CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
+            ],
         )
-        self.logger.info(
+        logger.info(
             f"Creating {df.shape[0]} sequencing-experiment-genomic-file "
-            f"links in {DATASERVICE_URL}"
+            f"links in {settings.DATASERVICE_URL}"
         )
         _ = self._load_entities(SEQ_EXP_GEN_FILE, df)
 
-        return se_gf_df
+        return df
 
     def _prep_harmonized_gf_df(self, df):
         """
         Part of Step 1
 
-        Clean up the genomic file DataFrame (output from get_s3_file_metadata)
-        and get it ready to be ingested by transforming column values and
-        standardizing the column names.
+        Clean up the genomic file DataFrame (_manifest_df_ joined with S3
+        data obtained from utils.scrape_s3) and get it ready to be ingested
+        by transforming column values and standardizing the column names.
         """
-        self.logger.info("Preparing genomic file DataFrame for ingest ...")
+        logger.info("Preparing genomic file DataFrame for ingest ...")
 
-        df['Hashes'] = df.apply(lambda row: {"ETag": row['ETag']}, axis=1)
-        df['urls'] = df.apply(lambda row: [row['Filepath']], axis=1)
-        df['file_type'] = df.apply(
-            lambda row: genomic_file_ext(row["Filename"]), axis=1,
+        df["Hashes"] = df.apply(lambda row: {"ETag": row["ETag"]}, axis=1)
+        df["urls"] = df.apply(lambda row: [row["Filepath"]], axis=1)
+        df["file_type"] = df.apply(
+            lambda row: genomic_file_ext(row["Filename"]),
+            axis=1,
         )
         col_rename = {
-            'KF Biospecimen ID': CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
-            'Data Type': CONCEPT.GENOMIC_FILE.DATA_TYPE,
-            'Filename': CONCEPT.GENOMIC_FILE.FILE_NAME,
-            'Hashes': CONCEPT.GENOMIC_FILE.HASH_DICT,
-            'Size': CONCEPT.GENOMIC_FILE.SIZE,
-            'urls': CONCEPT.GENOMIC_FILE.URL_LIST,
-            'file_type': CONCEPT.GENOMIC_FILE.FILE_FORMAT,
-            'Filepath': CONCEPT.GENOMIC_FILE.ID,
-            'Source Read': CONCEPT.GENOMIC_FILE.SOURCE_FILE,
+            "KF Biospecimen ID": CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
+            "Data Type": CONCEPT.GENOMIC_FILE.DATA_TYPE,
+            "Filename": CONCEPT.GENOMIC_FILE.FILE_NAME,
+            "Hashes": CONCEPT.GENOMIC_FILE.HASH_DICT,
+            "Size": CONCEPT.GENOMIC_FILE.SIZE,
+            "urls": CONCEPT.GENOMIC_FILE.URL_LIST,
+            "file_type": CONCEPT.GENOMIC_FILE.FILE_FORMAT,
+            "Filepath": CONCEPT.GENOMIC_FILE.ID,
+            "Source Read": CONCEPT.GENOMIC_FILE.SOURCE_FILE,
         }
         df = df[list(col_rename.keys())]
         df.rename(columns=col_rename, inplace=True)
@@ -226,40 +214,41 @@ class GenomicDataLoader(object):
         as a new column to the genomic file DataFrame _genomic_df_.
         Return _genomic_df_
         """
-        self.logger.info(
+        logger.info(
             "Fetching harmonized genomic file KF IDs from ingest cache ..."
         )
         cache_dict = {
-            ast.literal_eval(key)['external_id']: value
+            ast.literal_eval(key)["external_id"]: value
             for key, value in ingest_cache["genomic_file"].items()
         }
         hash_df = pd.DataFrame(
             list(cache_dict.items()),
             columns=[
-                CONCEPT.GENOMIC_FILE.ID, CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID
+                CONCEPT.GENOMIC_FILE.ID,
+                CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
             ],
         )
         genomic_df = genomic_df.merge(
             right=hash_df,
             on=CONCEPT.GENOMIC_FILE.ID,
-            how='inner',
+            how="inner",
         )
         return genomic_df
 
-    def _get_seq_experiment_genomic_files(self, harmonized=False):
+    def _get_seq_experiment_genomic_files(self):
         """
         Part of Step 3
 
         Get related sequencing experiments for a study's
-        genomic files. Returns the DataFrame from step 4.
+        genomic files. Returns the DataFrame from step 5.
 
-        1. Get the study's genomic files
+        1. Get the study's unharmonized genomic files
         2. Get the study's sequencing-experiments
         3. Get the study's sequencing-experiment-genomic-file links
         4. Extract the KF IDs from step 1, 2, 3
         5. Join DataFrames from 1, 2, 3
         """
-        self.logger.info(
+        logger.info(
             "Fetching sequencing experiment info for source genomic files ..."
         )
 
@@ -268,7 +257,7 @@ class GenomicDataLoader(object):
             The links given by the DataService yield strings of the form
             /entity-type/kf_id. This function will return the kf_id.
             """
-            return link.split('/')[-1]
+            return link.split("/")[-1]
 
         dfs = {}
         endpoints = {
@@ -278,13 +267,15 @@ class GenomicDataLoader(object):
         }
 
         # Steps 1-3
-        filter_dict = {}
         for endpoint in endpoints:
-            if endpoint == GEN_FILES:
-                filter_dict = {"harmonized": harmonized}
+            filter_dict = (
+                {"is_harmonized": False} if endpoint == GEN_FILES else None
+            )
             dfs[endpoint] = utils.get_entities(
-                DATASERVICE_URL, endpoint,
-                self.study_id, filter_dict=filter_dict
+                settings.DATASERVICE_URL,
+                endpoint,
+                self.study_id,
+                filter_dict=filter_dict,
             )
 
         # Step 4
@@ -292,7 +283,7 @@ class GenomicDataLoader(object):
         gdf = dfs[GEN_FILES][["kf_id", "external_id"]].rename(
             columns={
                 "kf_id": "gf_kf_id",
-                "external_id": CONCEPT.GENOMIC_FILE.SOURCE_FILE
+                "external_id": CONCEPT.GENOMIC_FILE.SOURCE_FILE,
             }
         )
         # Sequencing experiments
@@ -320,9 +311,8 @@ class GenomicDataLoader(object):
             lambda link: get_kf_id(link)
         )
         sgdf[CONCEPT.SEQUENCING.TARGET_SERVICE_ID] = sgdf[
-            "_links.sequencing_experiment"].apply(
-            lambda link: get_kf_id(link)
-        )
+            "_links.sequencing_experiment"
+        ].apply(lambda link: get_kf_id(link))
         # Merge se-gf with se
         sgdf = pd.merge(sedf, sgdf, on=CONCEPT.SEQUENCING.TARGET_SERVICE_ID)
 
@@ -339,15 +329,17 @@ class GenomicDataLoader(object):
     def _load_entities(self, entity_type, df):
         """
         Uses the ingest lib's load stage to load the entities of _entity_type_
-        from DataFrame _df_ into the Data Service
+        from DataFrame _df_ into the Data Service.
         """
         self.loader = LoadStage(
             self.target_api_cfg,
-            DATASERVICE_URL,
+            settings.DATASERVICE_URL,
             [entity_type],
             self.study_id,
             cache_dir=os.getcwd(),
         )
+        self.loader.logger = logger
+        self.loader.logger.setLevel(logging.DEBUG)
         self.loader.run({entity_type: df})
         return self.loader.uid_cache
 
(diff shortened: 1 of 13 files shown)


creator/ingest_runs/manifests.py
```diff
@@ -0,0 +1,295 @@
...
+from d3b_utils.aws_bucket_contents import fetch_bucket_obj_info
+from kf_lib_data_ingest.app import settings
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.common.io import read_df
+from kf_lib_data_ingest.etl.load.load import LoadStage
+from kf_utils.dataservice.scrape import yield_entities
+
+from urllib.parse import urlparse
+
+DATASERVICE_URL = "https://kf-api-dataservice.kidsfirstdrc.org/"
```
znatty22, line 14: @gsantia I think this URL will be different depending on the environment we're deploying to. So I think you will want to add this variable to each of the dev, testing, production settings files. The value should be read in from an environment variable. If not set the default should point to `LOCALHOST`. Don't want to accidentally modify prod

This shouldn't matter for your unit tests since you're mocking data service responses out. But it will matter later for deployment and also if we want to do integration tests with our data service deployments.

Later on we will need to update the deployment environment variables on S3 for each study creator deployment environment. I think when we deploy to the dev environment the URL would be set to localhost. When we deploy to QA we'd point the environment variable to QA data service and when we deploy to prod we'd point the var to production data service.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,482 @@
...
+            "participant": "PT",
+            "biospecimen": "BS",
+            "genomic_file": "GF",
+        }
+
+    @property
+    def study_bucket(self):
+        """ Return study bucket from study KF ID """
+        return (
+            f"kf-study-us-east-1-dev-"
```
gsantia, line 85: this should change based on where it's deployed, right?

znatty22: I don't this matters anymore since the tests won't actually interact with S3 bc of the mocks

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,356 @@
...
+            consoleHandler.setFormatter(logging.Formatter(format_))
+            root.setLevel("INFO")
+            root.addHandler(consoleHandler)
+
+        self.logger = logging.getLogger(type(self).__name__)
+
+        self.loader = None
+        self.target_api_cfg = ingest_settings.load().TARGET_API_CONFIG
+
+    def ingest_gwo(self, manifest_df):
```
dankolbman, line 49: Looking through this function, I can see why testing is going to be headache.
Consider all the dependencies needed to test this single function. From what I can see, you need:
1) A dataframe with the right data (It's difficult for me to figure out what that is)
2) Some sort of s3 buckets need to be setup in the correct manner
3) Genomic files need to be put in the s3 buckets
4) A manifest that references the genomic files needs to be placed in the s3 buckets
5) A dataservice that is populated with the required genomic files
6) A dataservice that is populated with the required sequencing experiments
7) A dataservice that relates genomic files to sequencing experiments as needed

When I run into a piece of code that requires such a vast amount of dependencies, it usually tells me that we can find better abstractions. If we can formulate such abstractions such that we only need to have one or two dependencies in place, it's going to make our life much easier when it comes to testing. Another benefit may be that we get some handy abstractions that we may use in other places.

If I may hint at what sort of abstractions may help here, let me first suggest considering how this may be approached _without_ the use of DataFrames and pandas. DataFrames are themselves useful abstractions because they allow quick analysis of data by establishing very general concepts about what you can do with data. This allows someone in an interactive environment to do common data operations such as reshape, filter, and add fields on a huge space of different data values. This is extremely powerful when the need is to explore and analyze data quickly, but this isn't quite the same need that we have. In fact, the need we have here is quite the opposite.

Here, we are trying to take a pretty rigid list of data (the GWO) and map it into another rigid datamodel (the genomic files and such in the Dataservice). It may seem handy that DataFrames offer batch operations to reshape data, but I would suggest that modeling and operating on our data on a per-instance basis is much closer to how it exists in reality (the dataservice is implemented in this fashion after all).

Consider this design:
If instead of a generic dataframe, you had a list of `GenomicFile`s. Each `GenomicFile` would have your standard properties that you see in Dataservice. There are then properties that depend on other properties such as how `size` needs `url` to resolve what the size of the file is. This 'size' field would probably have some implementation that looks at the protocol of the `url` and uses the appropriate resolver to find the value to return.

Now let's consider what we'd have to do to test this design:
First, we would probably want to  test that our `size` field returns the correct value. We would write `test_genomic_file_size()` where we would make sure this gets done right for each protocol. If we wanted to further this abstraction, we could write `FileResolver`s or similar that get invoked based on the protocol and we would then want to test each `FileResolver`. Writing these tests would require that we model each dependency and simulate the different possible results. For the `HTTPFile`, say, we just need to check what happens when the file gets resolved and what happens when a 404 is returned. The beauty here is that we only have this one straightforward dependency to mock.
As far as other tests and mocking, it may at first seem like we will need to mock out these resolvers whenever we want to access the `size` property so we don't try to make network requests during the test. However, because of the work we did to formulate the `FileResolver` abstraction,  it is almost trivial to implement something like a `SelfResolver` which stores the size on the `GenomicFile` and thus no network requests would be needed. We would then be able to use this resolver in all our tests and then have zero required dependencies when we test other features utilizing `GenomicFile`s.

One last point: I see that there does appear to be a requirement to provide a `DataFrame` to some of the ingest library functions. Perhaps this may seem to promote using `DataFrame`s more, but recall that we can easily adjust our abstraction to handle this. For example, perhaps we add a `to_df()` class method or similar to make that conversion just in time before the hand-off to ingest lib.

So finally, I hope this shows that re-framing to a more appropriate conceptual model can save tons of headache when testing. It should also go to show that it can allow new functionality to be added very rapidly (ie adding a new `FileResolver`). I think we could spend a bit of time to re-think the design here to set ourselves up for future work and to save ourselves from crazy testing hell.

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,356 @@
...
+        )
+        self.logger.info(
+            f"Creating {df.shape[0]} sequencing-experiment-genomic-file "
+            f"links in {DATASERVICE_URL}"
+        )
+        _ = self._load_entities(SEQ_EXP_GEN_FILE, df)
+
+        return se_gf_df
+
+    def _prep_harmonized_gf_df(self, df):
```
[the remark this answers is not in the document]
gsantia: This is a totally valid point. The file we used to ingest the unharmonized genomic files in the fake test study has a different format from the genomic workflow output manifests, so the same extract config can't be used again, but it should be possible to just write an extract config for this initial step of the process.

creator/ingest_runs/tasks.py
```diff
@@ -53,24 +55,28 @@ def check_gwo(version):
...
+    TODO docstring
+    """
+    versions = ingest_run.versions.all()
+    logger.info(
+        "Begin ingesting genomic workflow manifests: "
+        f"{len(versions)}: {pformat(list(versions))}"
     )
-    import time
+    rows = []
+    for version in ingest_run.versions.all():
```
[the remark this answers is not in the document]
gsantia: That check happens in the `run_ingest` task above

creator/ingest_runs/genomic_data_loader.py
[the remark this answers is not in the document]
znatty22: We might still be able to use an extract config to perform these transformations by making use of the ingest lib's ExtractStage here. Idk how hard or easy that will be / if changes are be needed in the ingest library since the ExtractStage wasn't originally designed to be a standalone utility.

Regardless of how the transformations happen here, I don't think it will affect our use of the ingest cache? Unless I'm missing something

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,356 @@
...
+LOCALHOST = "http://localhost:5000"  # testing
+DATASERVICE_URL = LOCALHOST  # During testing
+
+
+class GenomicDataLoader(object):
+
+    def __init__(self, study_id, init_logger=False):
+        self.study_id = study_id
+
+        if init_logger:
```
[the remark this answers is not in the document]
znatty22: Logger setup was included here bc we were running the GenomicDataLoader standalone in a python script / outside of the study creator for rapid dev purposes.

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
znatty22: I believe that will never happen since this is checked before this method is called up in `run_ingest`

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,572 @@
...
+            if endpoint == "genomic-files":
+                params["harmonized"] = False
+            resp = requests.get(base_url, params=params)
+            total = resp.json()["total"]
+            assert total == count, (
+                f"{endpoint} expected count {count} != {total} found"
+            )
+
+
+def delete_entities(dataservice_url, study_id=None, logger=None):
```
znatty22, line 515: This feature will be added to kf-utils-python once https://github.com/kids-first/kf-utils-python/pull/24 merges. We should remove this function from here and import it from kf-utils-python once that PR merges.

tests/integration/test_study_generator.py
```diff
@@ -0,0 +1,89 @@
...
+    delete_entities,
+)
+from tests.integration.fixtures import (
+    LOCAL_DATASERVICE_URL,
+    test_study_generator,
+    dataservice_setup,
+)
+
+
+def test_delete_entities():
```
znatty22, line 16: Remove this once the `delete_entities` function is removed from StudyGenerator since it is being implemented in https://github.com/kids-first/kf-utils-python/pull/24

znatty22: @gsantia This test module can be removed since study generator isn't part of the pr anymore

creator/ingest_runs/genomic_data_loader.py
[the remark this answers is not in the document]
znatty22: @gsantia I think its ok to get rid of this init_logger stuff

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+import pandas as pd
+import pytest
+
+
+@pytest.fixture()
+def study_id():
+    return "SD_YE0WYE0W"
+
+
+@pytest.fixture()
```
znatty22, line 19: Should probably be a module scoped fixture so that we're not running ingest_study every time this fixture is used in a test

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+    loader = GenomicDataLoader(
+        study_id=study_generator.study_id,
+        init_logger=True
+    )
+    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
+    loader.ingest_gwo(manifest_df)
+    mock_s3_scrape.assert_called_once()
+    mock_load_entities.assert_called_once()
+    entity_type, df = mock_load_entities.call_args_list[0][0]
+    assert type(df) == pd.DataFrame
```
znatty22, line 173:
```suggestion
    assert isinstancce(df, pd.Dataframe)
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+@pytest.fixture
+def mock_validate(mocker):
+    mock = mocker.patch(
+        "creator.ingest_runs.genomic_data_loader.GenomicDataLoader."
+        "_validate_load"
+    )
+    return mock
+
+
+def test_load_harmonized_genomic_files(
```
znatty22, line 156: Since you're only testing loading of the genomic files here, you probably don't need all of the mocks. You could just do with the mocks that would be called in `load_harmonized_genomic_files`. In your test here you can change `loader.ingest_gwo(manifest_df)` to `df = loader.load_harmonized_genomic_files(manifest_df)`

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+        CONCEPT.GENOMIC_FILE.ID,
+        CONCEPT.GENOMIC_FILE.SOURCE_FILE,
+        CONCEPT.GENOMIC_FILE.HARMONIZED,
+        CONCEPT.GENOMIC_FILE.VISIBLE,
+    }
+    for column in expected_columns:
+        assert column in df
+    assert df.shape[1] == len(expected_columns)
+
+    mock_load_bs_gf_links.assert_called_once()
```
znatty22, line 196: Similar to the comment above, since you're only testing the one load function you don't need to check this stuff here. This would be done in another test (like the one below)

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+    assert mock_load_entities.call_count == 2
+    # We already checked that the first call to mock_load_entities had the
+    # appropriate arguments in the previous test, and they won't change. So
+    # we can just do rudimentary checks here.
+    entity_type1, df1 = mock_load_entities.call_args_list[0][0]
+    assert entity_type1 == "genomic_file"
+    assert type(df1) == pd.DataFrame
+    assert not df1.empty
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
+    assert manifest_df.shape[0] == df1.shape[0]
```
znatty22, line 234: Similar to the comment above, I think some of this test is a little redundant bc you already tested/checked everything above this line in the previous test. Since you're only testing the `load_specimen_harmonized_gf_links` method you could probably replace the above with: 
```
    loader = GenomicDataLoader(
        study_id=study_generator.study_id, init_logger=True
    )
    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
    gfs = loader.load_harmonized_genomic_files(manifest_df)
    bsgfs = loader.load_specimen_harmonized_gf_links(gfs)
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+        CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
+        CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
+        CONCEPT.BIOSPECIMEN.ID,
+        CONCEPT.BIOSPECIMEN_GENOMIC_FILE.VISIBLE,
+    }
+    for col in expected_columns:
+        assert col in df2
+    assert df2.shape[1] == len(expected_columns)
+
+    # Finally check load_seq_exp_harmonized_genomic_files was called with the
```
znatty22, line 253: Similar to a previous comment. This stuff should probably in its own test for the `load_seq_exp_harmonized_genomic_files` method where the test setup would simply be:

```
def test_load_seq_exp_harmonized_genomic_files(
    study_generator, mock_s3_scrape, mock_load_entities, 
):
    loader = GenomicDataLoader(
        study_id=study_generator.study_id, init_logger=True
    )
    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
    df = loader.ingest_gwo(manifest_df)
    # All the checks you have below this line would be heree
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
+    loader.ingest_gwo(manifest_df)
+    mock_s3_scrape.assert_called_once()
+    assert mock_load_entities.call_count == 2
+    # We already checked that the first call to mock_load_entities had the
+    # appropriate arguments in the previous test, and they won't change. So
+    # we can just do rudimentary checks here.
+    entity_type1, df1 = mock_load_entities.call_args_list[0][0]
+    assert entity_type1 == "genomic_file"
+    assert type(df1) == pd.DataFrame
```
znatty22, line 231:
```suggestion
    assert isinstancce(df, pd.Dataframe)
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+    assert not df1.empty
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
+    assert manifest_df.shape[0] == df1.shape[0]
+
+    # Now check the biospecimen gf links loading
+    entity_type2, df2 = mock_load_entities.call_args_list[1][0]
+    assert entity_type2 == "biospecimen_genomic_file"
+    assert type(df2) == pd.DataFrame
```
znatty22, line 239:
```suggestion
    assert isinstancce(df, pd.Dataframe)
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,396 @@
...
+        CONCEPT.GENOMIC_FILE.ID,
+        CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
+        CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
+        CONCEPT.BIOSPECIMEN.ID,
+        CONCEPT.BIOSPECIMEN_GENOMIC_FILE.VISIBLE,
+    }
+    for col in expected_columns2:
+        assert col in df2
+    assert df2.shape[1] == len(expected_columns2)
+    # Check sequencing-experiment-genomic-files for the harmonized
```
znatty22, line 370: Similar to a previous comment. This stuff should probably in its own test for the `load_seq_exp_harmonized_genomic_files` method where the test setup would simply be:

```
def test_load_seq_exp_harmonized_genomic_files(
    study_generator, mock_s3_scrape, mock_load_entities, 
):
    loader = GenomicDataLoader(
        study_id=study_generator.study_id, init_logger=True
    )
    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
    df = loader.ingest_gwo(manifest_df)
    # All the checks you have below this line would be heree
```

znatty22 (changes requested):

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,350 @@
...
+GEN_FILES = "genomic-files"
+SEQ_EXP = "sequencing_experiment"
+SEQ_EXPS = "sequencing-experiments"
+SEQ_EXP_GEN_FILE = "sequencing_experiment_genomic_file"
+SEQ_EXP_GEN_FILES = "sequencing-experiment-genomic-files"
+BIO_GEN_FILE = "biospecimen_genomic_file"
+BIO_GEN_FILES = "biospecimen-genomic-files"
+LOAD_ENTITY_TYPES = {GEN_FILE, BIO_GEN_FILE, SEQ_EXP_GEN_FILE}
+
+LOCALHOST = "http://localhost:5000"
```
dankolbman, line 25: What is this used for? Can we use one of the settings instead?

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,350 @@
...
+DATASERVICE_URL = settings.DATASERVICE_URL
```
dankolbman, line 26: Is this necessary?

dankolbman (commented):
It looks like the tests are timing out on Circle, maybe in `tests/ingest_runs/test_genomic_data_loader.py`?

gsantia:
> It looks like the tests are timing out on Circle, maybe in `tests/ingest_runs/test_genomic_data_loader.py`?

Yeah, I'm trying to figure out why...

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        # Steps 1-3
+        filter_dict = {}
+        for endpoint in endpoints:
+            if endpoint == GEN_FILES:
+                filter_dict = {"harmonized": harmonized}
+            dfs[endpoint] = utils.get_entities(
+                DATASERVICE_URL, endpoint,
+                self.study_id, filter_dict=filter_dict
```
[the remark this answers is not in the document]
gsantia: Doing this will get all the genomic files. It'll still work but will potentially grab extra files and consequently links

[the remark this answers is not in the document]
znatty22: Oh I must have been looking at something old. I thought I saw `filter_dict={}` before.

tests/ingest_runs/test_genomic_data_loader.py
gsantia: We can't do this currently as-is because it relies on the `tmpdir` fixture which is scoped for functions

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,300 @@
...
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
+    assert manifest_df.shape[0] == df.shape[0]
+    # Check that the columns are as expected
+    for column in GF_EXPECTED_COLUMNS:
+        assert column in df
+    assert df.shape[1] == len(GF_EXPECTED_COLUMNS)
+
+
+def test_load_specimen_harmonized_gf_links(
+    study_generator, mock_s3_scrape, mock_load_entities, mock_load_se_gf_links,
```
znatty22, line 197:
```suggestion
    study_generator, mock_s3_scrape, mock_load_entities,
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,300 @@
...
+    mock_validate
```
znatty22, line 198:
```suggestion

```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,300 @@
...
+    )
+    return mock
+
+
+def test_load_harmonized_genomic_files(
+    mock_s3_scrape, study_generator, mock_load_entities
+):
+    """
+    Test the loading of the harmonized genomic files from the genomic workflow
+    output manifest. Mock the rest of the ingest process.
```
znatty22, line 177:
```suggestion
    output manifest. 
```

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,300 @@
...
+    assert df.shape[1] == len(GF_EXPECTED_COLUMNS)
+
+
+def test_load_specimen_harmonized_gf_links(
+    study_generator, mock_s3_scrape, mock_load_entities, mock_load_se_gf_links,
+    mock_validate
+):
+    """
+    Test the load_specimen_harmonized_gf_links function. This is the second
+    step of the ingestion process. Mock the steps after.
```
znatty22, line 202:
```suggestion
    step of the ingestion process.
```

znatty22 (changes requested):
Tests look good, just a few minor requests

creator/analyses/analyzer.py
```diff
@@ -3,6 +3,7 @@
 import xlrd
 from collections import defaultdict
 from functools import partial
+from graphql import GraphQLError
```
znatty22, line 6: I feel like this bug was fixed in a previous PR already merged into master?

tests/integration/fixtures.py
```diff
@@ -0,0 +1,35 @@
...
+
+        def __init__(self, *args, **kwargs):
+            kwargs["working_dir"] = os.path.join(tmpdir, "output")
+            super().__init__(*args, **kwargs)
+            self.dataservice_url = LOCAL_DATASERVICE_URL
+
+    return TestStudyGenerator
+
+
+@pytest.fixture
```
znatty22, line 28: Could you delete this fixture since its not being used? I forgot to do it before!

znatty22 (changes requested):
The test you have written look good, but it looks like codecov is failing bc a few more small functions need to be unit tested: https://app.codecov.io/gh/kids-first/kf-api-study-creator/compare/502/diff

Looks like the ingest run util functions and the actual ingest gwo task top level function.

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,299 @@
...
+    loader = GenomicDataLoader(FAKE_STUDY)
+    manifest_df = study_generator.dataframes['gwo_manifest.tsv']
+    loader.load_harmonized_genomic_files(manifest_df)
+    mock_s3_scrape.assert_called_once()
+    mock_load_entities.assert_called_once()
+    entity_type, df = mock_load_entities.call_args_list[0][0]
+    assert isinstance(df, pd.DataFrame)
+    assert entity_type == GEN_FILE
+    assert not df.empty
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
```
znatty22, line 188: Sorry I missed this before but...shouldn't this drop duplicates on something that uniquely identifies the harmonized genomic files? In this test we want to know that every harmonized gf in the manifest got submitted to the Dataservice. Dropping duplicates on participant ID would potentially get rid of rows you don't want to get rid of. The manifest might have rows with duplicate participant IDs if each participant has multiple specimens or if each participant has single specimen with multiple seq files per specimen. 

Actually thinking about it some more...do we even want to modify the manifest_df at all before comparing it to df2? Maybe we should just be comparing manifest_df rows to df2 rows. If they're equal then that means the inner join between the S3 scrape and the manifest didn't lose any rows

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,299 @@
...
+    gfs = loader.load_harmonized_genomic_files(manifest_df)
+    loader.load_specimen_harmonized_gf_links(gfs)
+    mock_s3_scrape.assert_called_once()
+    assert mock_load_entities.call_count == 2
+    # Now check the biospecimen gf links loading
+    entity_type2, df2 = mock_load_entities.call_args_list[1][0]
+    assert entity_type2 == BIO_GEN_FILE
+    assert isinstance(df2, pd.DataFrame)
+    assert not df2.empty
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
```
znatty22, line 214: Same comment as above. In this test we want to check that every harmonized gf got linked to one unique specimen (since these harmonized files represent a re-aligned version of the single specimen source sequencing file). So if we have 30 harmonized files then we should have 30 unique bs-gf links

tests/ingest_runs/test_genomic_data_loader.py
```diff
@@ -0,0 +1,299 @@
...
+    mock_s3_scrape.assert_called_once()
+
+    # Check that _load_entities_ was called the right number of times
+    assert mock_load_entities.call_count == 3
+    # Check sequencing-experiment-genomic-files for the harmonized
+    # files are loaded properly
+    entity_type, df = mock_load_entities.call_args_list[2][0]
+    assert entity_type == SEQ_EXP_GEN_FILE
+    assert isinstance(df, pd.DataFrame)
+    manifest_df = manifest_df.drop_duplicates(["KF Participant ID"])
```
znatty22, line 277: Same comment as above. We want to check that every harmonized gf got linked to one unique sequencing experiment.

znatty22 (changes requested):

tests/ingest_runs/test_genomic_data_loader.py
znatty22: These tests might all be working as is bc the test manifest_df might be inaccurate (my fault sorry 😬 ). If so, we should fix that.

gsantia: Yeah... This is a point that I've been confused about before.

znatty22: Here's a snippet of the manifest df with the s3 paths truncated for visual purposes:
image

This looks correct to me. We have a participant with one specimen that went through sequencing which produced 3 source/unharmonized files (1 cram, 1 crai, and 1 md5). The BIX team harmonized this and produced 1 harmonized file, the gVCF. The 3 rows are there because some information is repeated since there are 3 different source sequencing filles. So I think everything I said above still stands. You always want to do the "checking" with respect to the harmonized files since that's what we're loading into the Dataservice.

creator/ingest_runs/tasks/ingest_run.py
```diff
@@ -1,12 +1,16 @@
...
+import pandas
 
 from creator.decorators import task
+from creator.analyses.analyzer import extract_data
 from creator.files.models import FileType
+from creator.ingest_runs.genomic_data_loader import GenomicDataLoader
 from creator.ingest_runs.models import IngestRun
 
-
 logger = logging.getLogger(__name__)
```
znatty22, line 15: Not sure if you've experienced the same thing but I noticed in my validation branch that I needed to set the log level to INFO on my module logger in order to get the info statements from my task to show up in the logs. I think the default log level is set to WARN so the info statements get filtered out

tests/ingest_runs/test_genomic_data_loader.py
gsantia: Makes sense. So it seems that dropping duplicates on `Filepath` is what we want. I do think we need to drop rows for all these tests though. This manifest has 30 rows and 10 harmonized genomic files. So then we'd have 10 biospecimen-genomic-file links and 10 genomic-file-sequencing-experiment links as well.

tests/ingest_runs/test_genomic_data_loader.py
gsantia: I agree. But we only have 10 harmonized files, right?

znatty22 (approved):
Looks ok but the circleci status check seems to be hanging

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,348 @@
...
+
+        Extract the genomic_file KF IDs from the ingest cache and add the IDs
+        as a new column to the genomic file DataFrame _genomic_df_.
+        Return _genomic_df_
+        """
+        logger.info(
+            "Fetching harmonized genomic file KF IDs from ingest cache ..."
+        )
+        cache_dict = {
+            ast.literal_eval(key)["external_id"]: value
```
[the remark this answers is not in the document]
gsantia: It's a `defaultdict` where the keys are the various entity types, and the values are dicts representing the entries for those entity types which were ingested. In these dicts, the key is a string representation of a dictionary which uniquely identifies the entity in question, and the value is the kf_id the ingest library assigns that entity.

An example of a key-value pair for a genomic file entity:
```
"{'study_id': 'SD_ME0WME0W', 'external_id': 's3://kf-study-us-east-1-dev-sd-me0wme0w/source/genomic-files/genomic-file-0-0.cram'}": 'GF_7E6SNVFR
```

dankolbman (approved):
# 🐕

PR #506, opened 2020-11-05
✨ Add ingest run jobs

### LINKED ISSUE #500: Add ingest run jobs

gsantia:
Closes #500.

Implements the ingest run job lifecycle for running an ingest task and canceling it (`tasks.run_ingest` and `tasks.cancel_ingest`). This includes adding a `state` field to `IngestRun` in order to capture the job's state. Also adds events to help track start, cancel, and failure of ingest runs.

Sets up the architecture for ingest jobs to occur, but the actual ingestion of the genomic workflow manifests will be found in #502 .


creator/ingest_runs/models.py
```diff
@@ -84,25 +89,43 @@ def compute_name(self):
...
+        - Before a file version is deleted and that version is involved in
+          an IngestRun with an active job
+
+        TODO:
+        After we figure out how to kill job
+        See https://github.com/rq/rq/issues/684
         """
         pass
 
+    @staticmethod
```
znatty22, line 120: @Gio can you change this use the `@property` decorator and just call it `queue`, so that we can access like this: `IngestRun.queue` instead of like this `IngestRun.get_job_queue()`

creator/ingest_runs/models.py
```diff
@@ -1,13 +1,17 @@
...
 from django.contrib.auth import get_user_model
+import django_rq
+from rq.job import NoSuchJobError, JobStatus
 
 from creator.files.models import Version
 from creator.jobs.models import JobLog
 
 DELIMITER = "-"
 NAME_PREFIX = "INGEST_REQUEST"
+INGEST_QUEUE_NAME = 'ingest'
```
znatty22, line 14: @Gio can you move this into the settings files and then import it in this module

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+from creator.files.models import FileType
+from creator.ingest_runs.models import IngestRun
+from creator.utils import stop_job
+
+
+logger = logging.getLogger(__name__)
+
+
+@task("run_ingest")
+def run_ingest(ingest_run_id=None):
```
znatty22, line 16: In all of our calls to `run_ingest` and `cancel_ingest` we pass in `str(ingest_run.id)`. Maybe we should just pass the uuid in and then do the string conversion in `get_ingest_run` so that you don't have a bunch of calls to `str()`

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+    logging.info(f"Preparing ingest run {ingest_run.pk} for processing.")
+
+    # Check if any IngestRuns with the same input hash exist. If so, cancel
+    # them and delete the newly created IngestRun, which never gets started.
+    # If not, start a new IngestRun.
+    if cancel_duplicate_ingest_runs(ingest_run):
+        ingest_run.delete()
+    else:
+        # Update run state so we know it's running
+        logging.info(f"Start ingest {ingest_run.id}")
```
znatty22, line 30: Since the above log statement uses the `.pk` field, let's be consistent and use that in all of the log statements. And I like `.pk` better than `.id` anyway because if we later change the primary key to something other than `id` then we don't have to update these log statements.

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+        ingest_run.start()
+        ingest_run.save()
+
+        # Check that the files are genomic workflow manifests
+        versions = ingest_run.versions.all()
+        are_gwo = (check_gwo(version) for version in versions)
```
znatty22, line 36:
```suggestion
        are_gwo = all([check_gwo(version) for version in versions])
```

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+        if all(are_gwo):
```
znatty22, line 37:
```suggestion
        if are_gwo:
```

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+            logging.info(
+                "All files in the ingest run are genomic workflow manifests."
+            )
+            ingest_genomic_workflow_output_manifests(ingest_run)
+        else:
+            raise Exception("Unknown file type detected in the ingest run.")
```
znatty22, line 43: Before the exception, I think we need a `ingest_run.fail()` call

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+    others = IngestRun.objects.filter(
+        state="started",
+        input_hash=ingest_run.input_hash
+    ).all()
+    others_flag = others.count() > 0
 
+    # Cancel them
+    for running_ingest in others:
+        ingest_id = str(running_ingest.id)
+        logging.info(f"Canceling previous run '{ingest_id}'")
```
znatty22, line 63: You could remove this since you have a similar log statement inside `cancel_ingest_run`

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+
+
+def get_ingest_run(ingest_run_id):
+    """
+    Obtain the IngestRun with id _ingest_run_id_.
+    """
+    try:
+        ingest_run = IngestRun.objects.get(pk=ingest_run_id)
+    except IngestRun.DoesNotExist as err:
+        logging.info(
```
znatty22, line 104: Probably want a `ingest_run.fail()` here too

znatty22 (changes requested):
Lot of good work here! Reviewed the code but not the tests yet. Will do that in a second review.

creator/ingest_runs/tasks.py
gsantia: I don't follow this, say
`are_gwo = (True, False, True)`, then
bool(are_gwo) = True. We want every entry in `are_gwo` to be `True`.

creator/ingest_runs/tasks.py
gsantia: it works for generators too!

creator/ingest_runs/tasks.py
znatty22: I think you could remove this log statement because when the exception is raised it will get logged by the job logger/task decorator and have pretty much the same info in it.

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+        ingest_run.save()
+
+        # Check that the files are genomic workflow manifests
+        versions = ingest_run.versions.all()
+        are_gwo = (check_gwo(version) for version in versions)
+        if all(are_gwo):
+            logging.info(
+                "All files in the ingest run are genomic workflow manifests."
+            )
+            ingest_genomic_workflow_output_manifests(ingest_run)
```
znatty22, line 41: I think this line should probably be wrapped in a general try except so that we can catch anything that goes wrong fail the run, and then re-raise the exception to halt execution. This would handle anything unexpected that could happen during ingest, as well as job timeouts which cause an exception if the job duration takes longer than the default queue timeout.

znatty22 (changes requested):

creator/ingest_runs/tasks.py
gsantia: If there's an exception then there's no `ingest_run` to use

tests/ingest_runs/test_tasks.py
```diff
@@ -0,0 +1,156 @@
...
+
+    mock_enqueue.assert_called_once_with(cancel_ingest, str(old_ir.id))
+    mock_enqueue.reset_mock()
+    assert new_ir.state == "waiting"
+
+    with pytest.raises(Exception):
+        assert get_ingest_run(new_ir_id)
+
+    """
+    2) Non-GWO Case
```
znatty22, line 82: Probably want to check that the ingest run has a failed state too

znatty22 (changes requested):

tests/ingest_runs/test_signals.py
```diff
@@ -0,0 +1,98 @@
...
+    Test that the pre_delete signals are properly called and
+    function accordingly.
+    """
+    mock_cancel = mocker.patch(
+        "creator.ingest_runs.signals.cancel_invalid_ingest_runs"
+    )
+    # Create data
+    for _ in range(4):
+        prep_file(authed=False)
+    files = list(File.objects.all())
```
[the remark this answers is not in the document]
gsantia: It's a `QuerySet`. The reason I cast it to a list here is I don't think the order of the returned elements is guaranteed from call to call for `.objects`

tests/ingest_runs/test_tasks.py
```diff
@@ -0,0 +1,154 @@
...
+
+    new_ir = setup_ingest_run(happy_versions, user)
+    assert count_ingest_runs() == 3
+    run_ingest(new_ir.id)
+    # The above call should delete _new_ir_
+    assert count_ingest_runs() == 2
+
+    mock_enqueue.assert_called_once_with(cancel_ingest, old_ir.id)
+    mock_enqueue.reset_mock()
+    assert new_ir.state == "waiting"
```
znatty22, line 76: Shouldn't `new_ir` be deleted by this point? So probably no need to check its state

tests/ingest_runs/test_tasks.py
```diff
@@ -0,0 +1,154 @@
...
+
+    """
+    2) Non-GWO Case
+    Call run_ingest on an IngestRun with a version that doesn't have a GWO
+    root_file. Exception should be raised.
+    """
+    bad_ir = setup_ingest_run(file_versions, user)
+    with pytest.raises(Exception):
+        assert run_ingest(bad_ir.id)
+
```
znatty22, line 89: Once you add the try/except around the call to `ingest_genomic_worflow_manifest`, we should probably have a test that mocks that function and has an exception side effect so that we can check that the ingest_run goes to fail state

znatty22 (changes requested):
I just dropped a lot of review comments 😬 , but it looks good and is coming along well!

tests/ingest_runs/test_signals.py
```diff
@@ -0,0 +1,98 @@
...
+    This gives us the following IRs:
+    (1), (2), (3), (1, 2), (2, 3), (1, 3), (1, 2, 3)
+    """
+    ingests_with_v1 = []
+    for versions in powerset:
+        ingest = setup_ir(versions, user)
+        if v1 in versions:
+            ingests_with_v1.append(str(ingest.id))
+
+    assert count_ingest_runs() == 7
```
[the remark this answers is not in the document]
gsantia: `powerset` is an `itertools.chain` object with no `__len__` attr

creator/ingest_runs/tasks.py
znatty22: Oh oops, you're right

tests/ingest_runs/test_signals.py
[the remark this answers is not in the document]
znatty22: Oh I see. Its prob a generator. I think its still a good idea to check the length of the iterable rather than hard code it in case you change the number of runs later

tests/ingest_runs/test_signals.py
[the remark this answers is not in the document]
znatty22: Oh I see 👍

creator/ingest_runs/models.py
```diff
@@ -73,6 +82,39 @@ class Meta:
...
+    @transition(field=state, source="started", target="canceled")
+    def cancel(self):
+        """
+        Cancel the ingest run on request from the user or on account of a new
+        ingest run being started for the same input parameters.
+        """
+        stop_job(str(self.id), queue=IngestRun.get_job_queue(), delete=True)
+
+    @transition(field=state, source=["started", "canceled"], target="failed")
+    def fail(self):
```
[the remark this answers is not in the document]
znatty22: The only time a run will be in waiting state is when you create it which happens here: https://github.com/kids-first/kf-api-study-creator/blob/1f5b459ea1d3896bb42da928d36856ea4732db41/creator/ingest_runs/mutations.py#L57. Then it will immediately change to started state when the start ingest task executes. With the way things work right now I don't think there is a way for the frontend to cancel the run when its in waiting state. 

But I guess it doesn't hurt to add "waiting" to this list here and that would set us up for a future where a user could queue up a ingest run but then decide to cancel it before they start it.

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,109 @@
...
+    """
+    Run ingest process for a given set of files.
+    """
+    ingest_run = get_ingest_run(ingest_run_id)
+    logging.info(f"Preparing ingest run {ingest_run.pk} for processing.")
+
+    # Check if any IngestRuns with the same input hash exist. If so, cancel
+    # them and delete the newly created IngestRun, which never gets started.
+    # If not, start a new IngestRun.
+    if cancel_duplicate_ingest_runs(ingest_run):
```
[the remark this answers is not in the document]
znatty22: Hmm yea, maybe this check should happen in start ingest run mutation. We can check if an ingest run with the input hash exists there. If yes, we queue cancel_ingest task. If no, we create new ingest run and queue run_ingest task (what we have now). @gio thoughts?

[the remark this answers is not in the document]
gsantia: I wish that was my github name...

creator/ingest_runs/signals.py
```diff
@@ -9,7 +17,68 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    )
+    cancel_invalid_ingest_runs(instance)
+
+
+def cancel_invalid_ingest_runs(version):
+    """
+    Cancel all currently running IngestRuns that contain _version_ among its
+    Versions.
+    """
+    invalid_ingests = IngestRun.objects.filter(
```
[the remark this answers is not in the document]
gsantia: I wasn't sure how to handle those. How might an IngestRun be left in that state? Some kind of error?

creator/ingest_runs/models.py
```diff
@@ -90,26 +132,14 @@ def compute_name(self):
         )
         return DELIMITER.join([NAME_PREFIX, version_id_str])
 
-    def start(self):
+    @staticmethod
+    def get_job_queue(**kwargs):
```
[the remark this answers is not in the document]
gsantia: I'm not sure. @znatty22 ?

tests/ingest_runs/test_signals.py
```diff
@@ -0,0 +1,98 @@
...
+        "creator.ingest_runs.signals.cancel_invalid_ingest_runs"
+    )
+    # Create data
+    for _ in range(4):
+        prep_file(authed=False)
+    files = list(File.objects.all())
+    f1 = files[0]
+    v1, v2, v3, v4 = [f.versions.first() for f in files]
+    # Make sure two of the Versions have the same root file
+    v2.root_file = f1
```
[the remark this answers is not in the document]
gsantia: I assigned `f1 = files[0]` here because `f1.delete()` gets called further below

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
znatty22: Oo oops! Wrong person lol. 

I think Dan was just talking about the call to ingest_run.delete()? We shouldn't create a new one if we know we are going to have to delete it. I think we still want to cancel the existing ingest run so that we can achieve the "start, stop, start" behavior when someone pushes the ingest button 3 times.

creator/ingest_runs/models.py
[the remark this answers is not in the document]
znatty22: Eh its not _needed_. It really could just be a property method on IngestRun. I made it a get method to make it obvious you can supply kwargs to it. The point of having something was so that we don't have to hard code the name of the ingest queue in a bunch of places. Doesn't matter much to me though. We can make it a property method? Or put it in the settings file?

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
gsantia: What I meant is, if we have a check in the start ingest run mutation for an ingest run with the same input hash, then we can't possibly have any ingest runs with duplicate input hashes.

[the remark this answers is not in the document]
znatty22: I think its still needed though right? Bc thats the only place where an ingest run is canceled inside the cancel_ingest task. I guess it should just be called `cancel_prev_ingest_run`

creator/ingest_runs/models.py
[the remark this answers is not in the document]
dankolbman: So this is equivalent to `partial(get_queue, name=INGEST_QUEUE_NAME)`?

It seems a lot more transparent to just write out `get_queue(INGEST_QUEUE_NAME)` or have a property `.queue` since I don't think we use the other kwargs anyway.

creator/ingest_runs/models.py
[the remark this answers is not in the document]
dankolbman: What if the queue is down or redis gets restarted or the job gets lost or the workers aren't consuming? I would think the user may want to cancel if it's been stuck in the waiting state for too long or perhaps some automated cleanup system would scan for things that have been in waiting for too long and cancel the long running ones.

creator/ingest_runs/signals.py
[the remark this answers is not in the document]
dankolbman: Wouldn't an IngestRun that's been created but not yet started be in a waiting state? It's possible that the `run_ingest` never gets run due to some failure after which you'd be left with an invalid IngestRun in the waiting state.

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
dankolbman: That sounds right to me.
Do you only care about unique ingest runs in the non-terminal states, tho?
With the way things are now, how would someone re-running an IngestRun work?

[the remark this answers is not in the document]
gsantia: I just realized, if we have the start ingest run mutation calls the `cancel_ingest` task as needed, what's the purpose of the cancel ingest run mutation? As of now, it's not being used, right?

creator/ingest_runs/signals.py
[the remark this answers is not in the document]
gsantia: You're right, I'll add cancellation of IngestRuns in the `waiting` state as well

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
znatty22: Yea for now there is no purpose. But I think later on as the UI evolves, the start ingest mutation would only start the ingest and some other button on the UI would call the cancel ingest mutation to stop the run

creator/ingest_runs/models.py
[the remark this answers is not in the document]
gsantia: Alright I've changed it to be a property `.queue` on the class instances. Needed to make some changes to how mocker was used in the tests to reflect this change.

creator/ingest_runs/tasks.py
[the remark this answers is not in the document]
gsantia: As an update, I've done this, but now the unit tests for this need to be rewritten

creator/ingest_runs/signals.py
```diff
@@ -9,7 +17,68 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+        state="started",
+        versions=version
+    ).all()
+    for ingest in invalid_ingests:
+        django_rq.enqueue(cancel_ingest, ingest.id)
+
+
+@receiver(post_transition, sender=IngestRun)
+def ingest_run_post_transition(sender, instance, name, source, target,
```
dankolbman, line 53: Is this black formatted?

creator/ingest_runs/tasks.py
```diff
@@ -1,13 +1,88 @@
...
+
+def get_ingest_run(ingest_run_uuid):
+    """
+    Obtain the IngestRun with uuid _ingest_run_uuid_.
+    """
+    ingest_run_id = str(ingest_run_uuid)
+    try:
+        ingest_run = IngestRun.objects.get(pk=ingest_run_id)
+    except IngestRun.DoesNotExist:
+        raise
```
dankolbman, line 86: Immediately raising is equivalent to not handling the exception. You could remove the try/except to get the same result.

dankolbman (changes requested):

creator/ingest_runs/tasks.py
znatty22: I think originally this try/except was here to catch the exception and then call `ingest_run.fail()` before reraising, but I don't think we need this method anymore since we have the blank try/except up on line 37 which does the same thing.

creator/ingest_runs/mutations.py
```diff
@@ -96,7 +115,9 @@ def mutate(self, info, id):
         except IngestRun.DoesNotExist:
             raise GraphQLError(f"IngestRun {obj_id} was not found")
 
-        django_rq.enqueue(cancel_ingest, args=(str(ingest_run.id),))
+        ingest_run.queue.enqueue(
+            cancel_ingest, args=(ingest_run.id,)
```
znatty22, line 119:
```suggestion
            cancel_ingest, args=(str(ingest_run.id),)
```

creator/ingest_runs/mutations.py
```diff
@@ -56,19 +55,39 @@ def mutate(self, info, input):
...
             ingest_run.save()
 
-        uuid_str = str(ingest_run.id)
-        django_rq.enqueue(
-            run_ingest,
-            args=(uuid_str,),
-            job_id=uuid_str,
-        )
-
+        # When IngestRuns are created, they're in the waiting state, so we
```
znatty22, line 62: Not sure where to put this so just putting it here. Could you do this instead:
```python
        # Create ingest run for a set of file versions, if one does not exist
        # with same input_hash
        ingest_run = IngestRun()
        input_hash = ingest_run.compute_input_hash(versions=versions)
        
        # Check if duplicate ingest runs exist 
        irs_with_same_input_hash = IngestRun.objects.filter(
            state__in=["waiting", "started"],
            input_hash=ingest_run.input_hash,
        ).all()

        if irs_with_same_input_hash.count() > 0:
            cancel_duplicate_ingest_runs(irs_with_same_input_hash)
        else:
            with transaction.atomic():
                ingest_run.save()
                ingest_run.creator = user
                ingest_run.versions.set(versions)
                ingest_run.save()

            ingest_run.queue.enqueue(
                run_ingest,
                args=(ingest_run.id,),
                job_id=str(ingest_run.id),
            )
        return StartIngestRunMutation(ingest_run=ingest_run)
```

This way you don't have to save the run to the db and then delete it if you find duplicates. Also, that ^ requires a slight mod to the `IngestRun.compute_hash` method:

```python
    def compute_input_hash(self, versions=None):
        """
        Compute an MD5 hash digest from the IngestRun's input parameters:
            - List of file version ids (Version.kf_id)
        """
        versions = versions or self.versions.all()
        version_id_str = "".join(sorted(v.kf_id for v in versions))
        return hashlib.md5(version_id_str.encode("utf-8")).hexdigest()
```

znatty22 (changes requested):

creator/ingest_runs/mutations.py
gsantia: ah yes this makes sense. i was trying to figure out how to do it without saving and deleting. but didn't think to change the `compute_hash` method

creator/ingest_runs/signals.py
gsantia: no, done

creator/ingest_runs/mutations.py
gsantia: Need to do 
```
# Check if duplicate ingest runs exist 
        irs_with_same_input_hash = IngestRun.objects.filter(
            state__in=["waiting", "started"],
            input_hash=input_hash,
        ).all()
```
instead and it works

creator/ingest_runs/models.py
```diff
@@ -141,6 +144,13 @@ def get_job_queue(**kwargs):
         kwargs["name"] = INGEST_QUEUE_NAME
         return django_rq.get_queue(**kwargs)
 
+    @property
```
znatty22, line 147: You can remove `get_job_queue` now that you have `queue` property

creator/ingest_runs/mutations.py
```diff
@@ -51,24 +51,46 @@ def mutate(self, info, input):
...
+        # with same input_hash
+        ingest_run = IngestRun()
+        input_hash = ingest_run.compute_input_hash(versions=versions)
+
+        # Check if duplicate IngestRuns exist
+        irs_with_same_input_hash = IngestRun.objects.filter(
+            state__in=["waiting", "started"],
+            input_hash=input_hash,
+        ).all()
+        print("all", IngestRun.objects.all())
```
znatty22, line 64: Probably want to 🔥 these prints?

creator/ingest_runs/models.py
gsantia: oops!

creator/ingest_runs/mutations.py
gsantia: oops again!

creator/ingest_runs/tasks.py
```diff
@@ -85,25 +66,21 @@ def ingest_genomic_workflow_output_manifests(ingest_run):
...
     while time.time() < t_end:
         time.sleep(1)
-        progress += '.'
+        progress += "."
         print(progress)
     print(f"Finished ingest run {ingest_run.id}")
 
 
-def get_ingest_run(ingest_run_id):
+def get_ingest_run(ingest_run_uuid):
```
znatty22, line 79: No need to convert the uuid to a str anymore since both the mutations already do that before they enqueue the tasks. And now I think you can just replace calls to this method with the one line thats in this method: ingest_run = `IngestRun.objects.get(pk=ingest_run_uuid)`

creator/ingest_runs/signals.py
```diff
@@ -9,7 +17,68 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    that include this Version.
+    """
+    logging.info(
+        f"Canceling all ingest runs containing version {instance.pk} before"
+        f" its deletion."
+    )
+    cancel_invalid_ingest_runs(instance)
+
+
+def cancel_invalid_ingest_runs(version):
```
znatty22, line 39: I don't see this method being used anywhere else? I think this functionality could just be called directly on line 36

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -87,10 +102,49 @@ def test_start_ingest_run(
         for version in ir["versions"]["edges"]:
             assert version["node"]["kfId"] in ir["name"]
 
+        # Check that the call to IngestRun.get_job_queue().enqueue occurs
```
znatty22, line 105:
```suggestion
        # Check that the call to IngestRun.queue.enqueue occurs
```

creator/ingest_runs/signals.py
gsantia: it's used in the signals tests

znatty22 (approved):
❇️ ❇️ ❇️ ❇️  I think its good to merge!

PR #509, opened 2020-11-11
✨ Kill rq jobs by id

znatty22:
Makes use of the new `send_kill_horse_command` to kill a job via job id

NOTE: This will likely go away once https://github.com/rq/rq/issues/1371 is implemented.

### DIFF
--- a/creator/utils.py
+++ b/creator/utils.py
@@ -9,7 +9,7 @@ logger = logging.getLogger(__name__)
 WAIT_FOR_FAIL_TIME = 0.25
 
 
-def stop_job(job_id, queue=None, delete=True):
+def stop_job(job_id, queue=None, delete=False):
     """
     Stop a running rq job and optionally delete the job from its queue
     and rq registries.
@@ -18,13 +18,11 @@ def stop_job(job_id, queue=None, delete=True):
     https://github.com/rq/rq/issues/1371 is implemented
 
     For jobs that have finished/failed, do nothing
-    For jobs that are queued/deferred cancel or delete them (see rq.job.Job)
+    For jobs that are queued/deferred remove them from their queues
     For jobs that have started, kill the worker horse executing the job
 
-    If delete is True, remove the job from its queue, all rq registries,
-    and from redis
-
-    If delete is False just remove the job from its queue
+    The delete boolean is a convenience option that when True will remove
+    the job from all registries and redis
 
     :param job_id: The id of the job: rq.Job.id
     :type job_id: str
@@ -47,29 +45,28 @@ def stop_job(job_id, queue=None, delete=True):
 
     status = redis_job.get_status()
 
-    # Skip jobs that have failed or are finished
+    # Do nothing for jobs that have failed or are finished
     if status in {JobStatus.FAILED, JobStatus.FINISHED}:
-        logger.info(f"Job {job_id} already complete, aborting")
-        return
+        logger.info(f"Job {job_id} already terminated: {status}")
 
     # Cancel jobs that haven't started yet
-    if status in {JobStatus.QUEUED, JobStatus.DEFERRED}:
+    elif status in {JobStatus.QUEUED, JobStatus.DEFERRED}:
         logger.info(
             f"Cancel job {job_id} by removing it from queue: {queue.name}"
         )
         redis_job.cancel()
-        return
-    # Kill running jobs
+
+    # Kill running job
     else:
         for worker in Worker.all(queue=queue):
-            print(worker.state, worker.get_current_job().id)
             if (worker.state == WorkerStatus.BUSY) and (
                 worker.get_current_job().id == job_id
             ):
                 logger.info(f"Killing job {job_id}")
                 send_kill_horse_command(queue.connection, worker.name)
+                break
 
-    # Delete job from failed job registry
+    # Optionally delete job from all rq job registries and redis
     if delete:
         time.sleep(WAIT_FOR_FAIL_TIME)
         logger.info(
--- a/creator/utils.py
+++ b/creator/utils.py
@@ -1,0 +1,79 @@
+import time
+import logging
+import django_rq
+from rq.job import NoSuchJobError, JobStatus, Job
+from rq.command import send_kill_horse_command
+from rq.worker import Worker, WorkerStatus
+
+logger = logging.getLogger(__name__)
+WAIT_FOR_FAIL_TIME = 0.25
+
+
+def stop_job(job_id, queue=None, delete=True):
+    """
+    Stop a running rq job and optionally delete the job from its queue
+    and rq registries.
+
+    NOTE: This method may not be needed after this
+    https://github.com/rq/rq/issues/1371 is implemented
+
+    For jobs that have finished/failed, do nothing
+    For jobs that are queued/deferred cancel or delete them (see rq.job.Job)
+    For jobs that have started, kill the worker horse executing the job
+
+    If delete is True, remove the job from its queue, all rq registries,
+    and from redis
+
+    If delete is False just remove the job from its queue
+
+    :param job_id: The id of the job: rq.Job.id
+    :type job_id: str
+    :param queue: The queue the job was submitted to
+    :type queue: rq.Queue
+    :param delete: whether to delete or cancel the job
+    :type delete: bool
+    """
+    logger.info(f"Preparing to stop redis job {job_id}")
+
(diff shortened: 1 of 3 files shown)


creator/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    if redis_job.get_status() == JobStatus.STARTED:
+        for worker in Worker.all(queue=queue):
+            if (
+                (worker.state == WorkerStatus.BUSY) and
+                (worker.get_current_job().id == job_id)
+            ):
+                logger.info(f'Killing job {job_id}')
+                send_kill_horse_command(queue.connection, worker.name)
+
+    # Wait until job has been moved to failed job registry to delete or
```
gsantia, line 67: So what would cause this to stall out for 30 seconds? And if that happens do we need to do anything afterwards?

znatty22: No it'll break out of the while loop when the status changes. See line 85

creator/utils.py
```diff
@@ -0,0 +1,85 @@
+import time
+import logging
+from rq.job import NoSuchJobError, JobStatus
+from rq.command import send_kill_horse_command
+from rq.worker import Worker, WorkerStatus
+
+WAIT_FOR_FAIL_TIMEOUT = 30
+
+
+def stop_job(job_id, queue=None, delete=True, logger=None):
```
dankolbman, line 10: What is the use case for passing a logger? This may cause unexpected behavior when trying to modify our logging setup.

creator/utils.py
```diff
@@ -0,0 +1,85 @@
...
+        queue = Queue(connection=Redis())
+
+    redis_job = None
+    try:
+        redis_job = queue.fetch_job(job_id)
+    except NoSuchJobError:
+        return
+
+    # Skip jobs that have failed or are finished
+    if (
```
dankolbman, line 50: Did you run this through the black formatter?

creator/utils.py
```diff
@@ -0,0 +1,85 @@
...
+            if (
+                (worker.state == WorkerStatus.BUSY) and
+                (worker.get_current_job().id == job_id)
+            ):
+                logger.info(f'Killing job {job_id}')
+                send_kill_horse_command(queue.connection, worker.name)
+
+    # Wait until job has been moved to failed job registry to delete or
+    # cancel job
+    while time.time() < (time.time() + WAIT_FOR_FAIL_TIMEOUT):
```
dankolbman, line 69: How long does this operation take and where do we plan to call it from?
The `while` is going to peg the CPU, possibly of a worker that might otherwise be serving requests, which could cause our autoscaler to kick up the instance count. It will also bombard redis with requests which can be a problem if we end up canceling lots of jobs at once.
Perhaps it would be better to queue another job to check up on the status ?

creator/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    :type delete: bool
+    """
+    if not logger:
+        logger = logging.getLogger(__name__)
+
+    logger.info(f"Preparing to stop redis job {job_id}")
+
+    # Find rq job
+    if not queue:
+        queue = Queue(connection=Redis())
```
dankolbman, line 41: Where does `Redis` come from?
I think `django_rq.get_queue()` may be preferable here.

tests/test_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    p.start()
+    worker.work(burst=True)
+    p.join(1)
+
+
+@pytest.mark.parametrize(
+    'delete',
+    [False, True]
+)
+def test_stop_job(queues, delete):
```
dankolbman, line 95: This seems like quite a complicated test setup. Think you could do the same with much less work by mocking out the response from `get_status()` and `fetch_job()` and also test more cases more easily and spend less time in the actual test execution.

creator/utils.py
znatty22: I switched it from `django_rq` bc I thought about submitting a PR to rq's repo but I don't think I'm going to do that anymore so I'll switch it back

creator/utils.py
znatty22: This operation will be called from the `IngestRun.cancel` method. Eh yea 30s was left from debugging. I think I'm just going to sleep for less than a second, maybe 0.1 bc I noticed during testing that's all it takes for rq to move the job to a registry. 

In general submitting another job to check on status seems like it would have the same problem bc the status might not have changed by the time you're inside the job. In this case it would probably work but I'd rather just sleep. Sigh the real solution is to have rq support subscription to job status changes. This issue looks like what we want: https://github.com/rq/rq/issues/717 but no ones looked at it in a while

creator/utils.py
znatty22: I'll remove it

tests/test_utils.py
znatty22: Hmm are those the bottlenecks? I think its my task which sleeps for 2 seconds. So the truth is I didn't want to deal with mocking things :) I just thought it would be easier but also more thorough to do real calls

creator/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        for worker in Worker.all(queue=queue):
+            if (worker.state == WorkerStatus.BUSY) and (
+                worker.get_current_job().id == job_id
+            ):
+                logger.info(f"Killing job {job_id}")
+                send_kill_horse_command(queue.connection, worker.name)
+
+    # Wait until job has been moved to failed job registry to delete or
+    # cancel job
+    time.sleep(WAIT_FOR_FAIL_TIME)
```
dankolbman, line 66: What's the impact of never calling `delete()` or `cancel()` on the job? It seems likely this could happen.

znatty22: If you don't call delete the killed jobs will pile up in the failed job registry. I guess it depends on if you care about that and if you want to differentiate between jobs that failed due to kill horse or something else.  

For cancel, I just realized I need to change the logic bc it removes jobs from the queue that haven't started. So I don't care about waiting for the failed job reg to update

gsantia (changes requested):

tests/test_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    if job_status in {JobStatus.QUEUED, JobStatus.DEFERRED}:
+        assert mock_job.cancel.call_count == 1
+
+    if job_status == JobStatus.STARTED:
+        assert mock_kill.call_count == 1
+
+    if delete:
+        mock_job.delete.call_count == 1
+
+    mock_job.reset_mock()
```
dankolbman, line 77: Is this needed? I was under the impression that all mocks should be reset between runs.

creator/utils.py
```diff
@@ -0,0 +1,79 @@
...
+        return
+    # Kill running jobs
+    else:
+        for worker in Worker.all(queue=queue):
+            print(worker.state, worker.get_current_job().id)
+            if (worker.state == WorkerStatus.BUSY) and (
+                worker.get_current_job().id == job_id
+            ):
+                logger.info(f"Killing job {job_id}")
+                send_kill_horse_command(queue.connection, worker.name)
```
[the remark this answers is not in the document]
znatty22: Oops ye

creator/utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Delete job from failed job registry
+    if delete:
+        time.sleep(WAIT_FOR_FAIL_TIME)
+        logger.info(
+            f"Delete job {job_id} by removing it from queue: "
+            f"{queue.name} and all job registries"
+        )
+        redis_job.delete()
```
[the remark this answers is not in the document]
znatty22: I don't think that can happen? If the execution gets down here it means the else block executed bc the job was running, and if its running then it has to be executing in one of the work horses.  

But that being said I wonder if the user should have the option to delete the job given the other cases where it either already failed/finished or we cancel it? Right now that doesn't happen because I return at the end of the other if blocks. Or I could just remove deletion entirely and leave it to the user to do that after calling this method?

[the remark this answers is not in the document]
znatty22: EDIT: In my last commit I made it so that delete=False by default. But if user sets it to True the job will always be deleted at the end of stop job. I change that though if we don't like it

tests/test_utils.py
znatty22: Ooo cool they do. I removed it

dankolbman (approved):
:+1:

gsantia (approved):

PR #510, opened 2020-11-16
♻️ Refactor settings

### LINKED ISSUE #504: Extract feature settings
Our feature flags (`FEAT_`) and general settings currently need to be repeated across all our settings files although they are similar in all environments. We should be able to generalize these within one `features.py` file to reduce these repetitions when adding or changing features.

dankolbman:
Moves common feature settings into their file 

Closes #504

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,0 @@
-DEFAULT_FILE_STORAGE=django_s3_storage.storage.S3Storage
-UPLOAD_DIR=source/uploads/
-AWS_S3_BUCKET_NAME=kf-study-us-east-1-dev-sd-me0owme0w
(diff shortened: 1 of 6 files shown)


znatty22 (commented):
There is a lot of repetitive code in all 3 settings files so can we also have a base settings file with shared settings that could be imported into development.py, testing.py, and production.py?

dankolbman:
I think this is the start to combining the settings files. I'm debating whether we want to consolidate to one and handle everything through the environment via django-environ, but I want to transition slowly to make sure things don't break across environments.

znatty22 (approved):
👍

znatty22:
> I think this is the start to combining the settings files. I'm debating whether we want to consolidate to one and handle everything through the environment via django-environ, but I want to transition slowly to make sure things don't break across environments.

I still vote for separate settings files as long as we can do something to DRY them up

dankolbman:
I still need to clean out EGO settings here now that it's removed.

PR #511, opened 2020-11-16
🔥 Remove ego authentication

dankolbman:
Completely removes ego authentication middleware.

### DIFF
--- a/creator/middleware.py
+++ b/creator/middleware.py
@@ -16,136 +16,11 @@ logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
 
-class EgoJWTAuthenticationMiddleware():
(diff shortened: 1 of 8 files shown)


znatty22 (commented):
I still see `ego` (e.g. ego_groups, ego_roles) in the following files:

```
creator/users/schema.py
creator/users/factories.py
creator/users/mutations.py
creator/models.py
creator/middleware.py
```
Do any of these files need to change?

dankolbman:
We still use `ego_groups` and `ego_roles` as user properties for legacy authentication. I want to remove/rename these, but I need to spend more time reviewing what impacts this will have.

znatty22 (approved):
👍

PR #513, opened 2020-11-17
✨ Add IngestRun state and events

### LINKED ISSUE #500: Add ingest run jobs

gsantia:
Partially closes #500 .


creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,31 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    Create a corresponding Event after the status of an IngestRun is changed.
+    """
+    TARGET_EVENT_TYPES = {
+        "running": "IR_STA",
+        "complete": "IR_COM",
+        "canceled": "IR_CAN",
+        "failed": "IR_FAI",
+    }
+    username = getattr(instance.creator, "display_name", "Anonymous user")
+    message = (
```
znatty22, line 33: Couple things about the message: 
1. Could you add the status into the message somewhere so the event timeline tells us what happened to the ingest ru?
2. It probably only makes sense to display the username in the message if the user started the ingest run or canceled it. It sounds weird to say something like "Gio changed the status of IngestRun to FAILED" haha

creator/ingest_runs/models.py
```diff
@@ -73,6 +82,35 @@ class Meta:
...
+        return
+
+    @transition(field=state, source="running", target="complete")
+    def complete(self):
+        """
+        Finish running the ingest process without error.
+        """
+        return
+
+    @transition(field=state, source="*", target="canceled")
```
znatty22, line 99: Hmm not sure about this but I think should only be able to cancel the IngestRun if its in running state because if its not in running state then there's nothing to cancel (the redis job would not even be created at this point)

creator/ingest_runs/models.py
```diff
@@ -73,6 +82,35 @@ class Meta:
...
+    def canceled(self):
+        """
+        Cancel the ingest run on request from the user or on account of a new
+        ingest run being started for the same input parameters.
+        """
+        return
+
+    @transition(field=state, source="*", target="failed")
```
znatty22, line 107: Also not sure about this but I think you can only get into failed state from running or canceled. The ingest run could be running and then unexpectedly fail or the ingest run could be canceled by the user causing it to go into failed state.

znatty22 (changes requested):

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,31 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    """
+    Create a corresponding Event after the status of an IngestRun is changed.
+    """
+    TARGET_EVENT_TYPES = {
+        "running": "IR_STA",
+        "complete": "IR_COM",
+        "canceled": "IR_CAN",
+        "failed": "IR_FAI",
+    }
+    username = getattr(instance.creator, "display_name", "Anonymous user")
```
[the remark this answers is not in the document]
gsantia: I'm not sure, I took this line from @dankolbman 's draft PR

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,38 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    """
+    TARGET_EVENT_TYPES = {
+        "running": "IR_STA",
+        "complete": "IR_COM",
+        "canceled": "IR_CAN",
+        "failed": "IR_FAI",
+    }
+    username = instance.creator.display_name or "Anonymous user"
+    if target in {"complete", "failed"}:
+        message = f"Status change of IngestRun {instance.id} with "
```
znatty22, line 35:
```suggestion
        message = f"IngestRun {instance.id} {target} for file versions {versions} "
```

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,38 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    else:
+        message = (
+            f"{username} changed the status of IngestRun {instance.id} "
```
znatty22, line 38:
```suggestion
            f"{username} {target} IngestRun {instance.id} for file versions {versions}"
```

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,38 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
         instance.input_hash = instance.compute_input_hash()
+
+
+@receiver(post_transition, sender=IngestRun)
+def ingest_run_post_transition(sender, instance, name, source, target,
+                               *args, **kwargs):
+    """
+    Create a corresponding Event after the status of an IngestRun is changed.
+    """
+    TARGET_EVENT_TYPES = {
```
znatty22, line 27:
```suggestion
    if target == running:
        target = 'started'
    TARGET_EVENT_TYPES = {
        "started": "IR_STA",
        "complete": "IR_COM",
        "canceled": "IR_CAN",
        "failed": "IR_FAI",
    }
    versions = instance.versions.all()
```

znatty22 (changes requested):
@gsantia Are you going to add the unit test for the events?

gsantia:
> @gsantia Are you going to add the unit test for the events?

Ya working on it

creator/ingest_runs/models.py
```diff
@@ -77,6 +85,35 @@ class Meta:
         help_text="The user who submitted this ingest run",
     )
 
+    @transition(field=state, source="waiting", target="running")
+    def started(self):
```
dankolbman, line 89: Is there a reason you chose `started` here? It seems more semantic to use the imperative here as you're telling the job to `start`, which transitions from the `waiting` state to the `running` state. The past participle doesn't seem as indicative to me.

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,41 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+
+
+@receiver(post_transition, sender=IngestRun)
+def ingest_run_post_transition(sender, instance, name, source, target,
+                               *args, **kwargs):
+    """
+    Create a corresponding Event after the status of an IngestRun is changed.
+    """
+    if target == "running":
+        target = "started"
```
dankolbman, line 28: Do you need to mutate the target here? Could you just map `running` onto `"IR_STA"` instead?

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,41 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+    TARGET_EVENT_TYPES = {
+        "started": "IR_STA",
+        "complete": "IR_COM",
+        "canceled": "IR_CAN",
+        "failed": "IR_FAI",
+    }
+    versions = list(instance.versions.all())
+    username = instance.creator.display_name or "Anonymous user"
```
dankolbman, line 36: This should be taken care of here: https://github.com/kids-first/kf-api-study-creator/blob/master/creator/models.py#L95

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,41 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+            f"{username} {target} IngestRun {instance.id} for file versions "
+            f"{versions} "
+        )
+    
+    first_version = instance.versions.first()
+    ev = Event(
+        study=first_version.study,
+        file=first_version.root_file,
+        version=first_version,
+        user=instance.creator,
```
dankolbman, line 52: You may want to consider excluding the user if the event is one that is not created in response to a request from the user.

creator/ingest_runs/models.py
gsantia: I did it to distinguish from the other `start` function that I had there. But I'm not sure if that function was going to be used. I assume it's just old boilerplate code I should delete

creator/ingest_runs/signals.py
gsantia: It's for the Event messages. Maybe we can change the values of `state`?

creator/ingest_runs/models.py
```diff
@@ -77,6 +85,35 @@ class Meta:
...
+
+    @transition(field=state, source="running", target="complete")
+    def complete(self):
+        """
+        Finish running the ingest process without error.
+        """
+        return
+
+    @transition(field=state, source="running", target="canceled")
+    def canceled(self):
```
znatty22, line 103: @Gio Similar to `started` you may want to delete the `cancel` placeholder and just call this method `cancel`

creator/ingest_runs/migrations/0001_initial.py
```diff
@@ -11,7 +11,6 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ("files", "0022_add_genomic_workflow_output_type"),
```
znatty22, line 14: I don't think this migration should have changed. Is this an unintentional change?

gsantia: This line appeared in the migration initially as an unintended side effect of providing a temporary default value for `id` when prompted by django. It had to be removed for the migrations to work

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -91,6 +92,35 @@ def test_start_ingest_run(
...
+        prep_file(authed=True)
+    version_ids = [
+        to_global_id("VersionNode", f.versions.first().kf_id)
+        for f in File.objects.all()
+    ]
+
+    # Start valid ingest runs for a batch of file versions
+    resp = send_query(client, START_INGEST_RUN, {"versions": version_ids})
+
+    assert Event.objects.count() == 1
```
znatty22, line 119: Right idea, but I a little extraneous since we already have a test for the start ingest run mutation. Could you move the event checking block into the `test_start_ingest_run` test in the `if allowed` block?

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -142,3 +172,35 @@ def test_cancel_ingest_run(
...
+        data={
+            "query": CANCEL_INGEST_RUN,
+            "variables": {
+                "id": to_global_id("IngestRunNode", ingest_run.id),
+            },
+        },
+        content_type="application/json",
+    )
+
+    assert Event.objects.count() == 1
```
znatty22, line 204: Same comment as above, except could you move this into the `test_cancel_ingest_run`

znatty22 (changes requested):

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -62,11 +63,14 @@ def send_query(client, query, input_dict):
     ],
 )
 def test_start_ingest_run(
-    db, clients, prep_file, mock_ingest_job, user_group, allowed
+    db, mocker, clients, prep_file, mock_ingest_job, user_group, allowed
 ):
     """
     Test the start ingest run mutation
     """
+    get_job_mock = mocker.patch(
```
znatty22, line 71:
```suggestion
    mock_ingest_queue = mocker.patch(
```

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -87,6 +91,10 @@ def test_start_ingest_run(
         for version in ir["versions"]["edges"]:
             assert version["node"]["kfId"] in ir["name"]
 
+        # Check that the call to IngestRun.get_job_queue().enqueue occurs
+        get_job_mock.assert_called_once()
```
znatty22, line 95:
```suggestion
        mock_ingest_queue.assert_called_once()
```

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -87,6 +91,10 @@ def test_start_ingest_run(
...
+        get_job_mock.reset_mock()
```
znatty22, line 96:
```suggestion
        mock_ingest_queue.reset_mock()
```

tests/ingest_runs/test_model.py
```diff
@@ -1,11 +1,18 @@
...
 from creator.files.models import File
 from creator.ingest_runs.models import IngestRun
 
+from django.contrib.auth import get_user_model
 
-def test_ingest_run(db, prep_file):
+User = get_user_model()
+
+
+def test_ingest_run(clients, db, prep_file):
```
znatty22, line 10: I think db has to come first since it sets up the test db, then clients

tests/ingest_runs/test_model.py
```diff
@@ -33,3 +40,38 @@ def test_ingest_run(db, prep_file):
...
+    assert event_type_num("IR_STA") == 3
+    assert event_type_num("IR_COM") == 1
+    assert event_type_num("IR_CAN") == 1
+    assert event_type_num("IR_FAI") == 1
+
+
+def setup_ir(file_versions, user):
+    ir = IngestRun()
+    ir.save()
+    ir.creator = user
```
znatty22, line 67: This line can go above the first save and then you can get rid of the subsequent save

gsantia: yeah i'm not sure which calls actually require a save

gsantia: which line?

creator/ingest_runs/models.py
```diff
@@ -77,6 +85,35 @@ class Meta:
...
     )
 
+    @transition(field=state, source="waiting", target="started")
+    def start(self):
+        """
+        Begin running the ingest process.
+        """
+        return
+
+    @transition(field=state, source="started", target="complete")
```
znatty22, line 95: Sorry plz don't kill me... small request. I think we should change it to  `target=completed` since all of the other states are in past tense.

creator/ingest_runs/signals.py
```diff
@@ -13,3 +16,38 @@ def ingest_run_pre_save(sender, instance, *args, **kwargs):
...
+
+@receiver(post_transition, sender=IngestRun)
+def ingest_run_post_transition(sender, instance, name, source, target,
+                               *args, **kwargs):
+    """
+    Create a corresponding Event after the status of an IngestRun is changed.
+    """
+    TARGET_EVENT_TYPES = {
+        "started": "IR_STA",
+        "complete": "IR_COM",
```
znatty22, line 29: Same comment here

znatty22 (changes requested):

znatty22 (approved):
💯 🎄 ✅

PR #515, opened 2020-11-20
🐛 Don't resolve downloadUrls if not enough info

dankolbman:
Does not resolve the `downloadUrl` for a version unless there is enough information to build it.


znatty22 (approved):
👍

PR #518, opened 2020-12-03
🔥 Remove image from email

dankolbman:
Removes the Kids First image from emails in attempt to avoid going to user's spam folder.


znatty22 (approved):
📭

PR #521, opened 2020-12-07
🔧 Ingest redis

gsantia:
Add ports to the redis service in `docker-compose.yml` to allow for testing with just the compose file.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 venv/
+__pycache__/
+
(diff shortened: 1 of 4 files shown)


znatty22 (changes requested):
@gsantia I thought you were making a PR to publish the default redis and postgres ports in the docker-compose file for general dev purposes? If so, I would point this PR to the master branch and include the postgres port (unless its already published?)

PR #522, opened 2020-12-07
:wrench: Add ports to redis and postgres services

gsantia:
Add appropriate ports to the redis and postgres services in `docker-compose.yml` to allow for testing with the compose stack.


dankolbman:
I'm a little cautious about using containers on the docker-compose network for testing because 1) often these ports are already being used on a developer's local which will cause docker-compose to fail 2) the services in docker-compose will often be storing local application state used during development and will be overwritten by tests.
Can we opt to use host-bound, test-specific containers to run tests on instead?

PR #525, opened 2020-12-15
✨ Admin configured banner messages

### LINKED ISSUE #517: Add banner messages
It would be useful to have banner messages that admins can add at any time that notify the users of any new features, changes, or problems in the Data Tracker.

The model should include some basic fields:
- creator
- createdAt
- active
- start/end date
- message
- type ('info', 'warning', 'error')

We should have queries to support:
- list all messages (filter by active, sort by createdAt)
- add new message
- change message (active, message, start/end date)

To communicate current messages out to the user, we may consider adding all active messages that are within the start/end date to the `status` query schema.

znatty22:
Closes #517 

**Change Summary**
- Refactor status app by moving settings/features GraphQL API into `creator/status/settings`
- Add banner message GraphQL API to status app in `creator/status/banners`


creator/status/banners/queries.py
```diff
@@ -0,0 +1,35 @@
...
+from creator.status.banners.nodes import BannerNode
+from creator.status.banners.models import Banner
+
+
+class BannerFilter(FilterSet):
+    order_by = OrderingFilter(fields=("created_on",))
+
+    class Meta:
+        model = Banner
+        fields = ["enabled"]
```
dankolbman, line 15: I suppose the main use of the banners query is to fetch all banners that should be displayed and, to do that, we'll also need to filter by the start date and the end date.
You could add those here, but graphene will add those arguments as `createdOn_Gt` and `createdOn_Lt` which are pretty ugly. I'd suggest doing it like we do with events:  https://github.com/kids-first/kf-api-study-creator/blob/544a4261176566fd45b05874b97ec578d5e01248/creator/events/schema.py#L32-L37

creator/status/banners/nodes.py
```diff
@@ -0,0 +1,29 @@
...
+        """
+        user = info.context.user
+
+        if not (user.has_perm("status.view_banner")):
+            raise GraphQLError("Not allowed")
+
+        try:
+            banner = cls._meta.model.objects.get(id=id)
+        except cls._meta.model.DoesNotExist:
+            raise GraphQLError("Banner was not found")
```
dankolbman, line 27: You can test this pretty easily by adding banners here: https://github.com/kids-first/kf-api-study-creator/blob/master/tests/test_not_found.py

creator/status/banners/queries.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Query(object):
+    banner = relay.Node.Field(BannerNode, description="Get a single banner")
+    banners = DjangoFilterConnectionField(
```
dankolbman, line 20: Can you rename this `allBanners` to be conformant with all our other list queries?

dankolbman (changes requested):
Looks good! Just a minor tweak or two to make it easier to use on the front end.

creator/status/banners/queries.py
znatty22: @dankolbman So I actually had it as `allBanners` before bc thats what the `startresource` cmd generated for me but I changed it because the other things within the Status query don't follow that naming convention (`settings`, `features`). Should I still change it to `allBanners`?

dankolbman (approved):
# :bacon:

PR #526, opened 2020-12-23
💬 Make extension map more strict

fiendish:
We don't want to accidentally match some weirdo ".obai" format or something


znatty22 (approved):
🎅

PR #527, opened 2020-12-23
♻️ Remove unnecessary forwarding lambdas

fiendish:
`lambda x: foo(x)` should be the same as `foo(x)` if foo doesn't depend on state


fiendish:
Failure looks unrelated?

znatty22 (approved):
🎁

PR #529, opened 2020-12-23
🐛 Don't use os.path.split on S3 keys

fiendish:
https://docs.python.org/3.8/library/os.path.html says
```
Note: Since different operating systems have different path name
conventions, there are several versions of this module in the standard
library. The os.path module is always the path module suitable for the
operating system Python is running on, and therefore usable for local
paths.
```

But since no other path methods are being used and S3 keys aren't really file paths anyway, it seems reasonable to just split on "/" instead of invoking os.posixpath.

I made a def instead of just a lambda because the filename is often useful for other extract operations as well.

### DIFF
--- a/creator/extract_configs/templates/s3_scrape_config.py
+++ b/creator/extract_configs/templates/s3_scrape_config.py
@@ -19,8 +19,6 @@ operations.append(
     )
 )
 """
-import os
-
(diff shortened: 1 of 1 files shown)


znatty22 (approved):
✅

PR #532, opened 2021-01-07
🔥 Remove dotenv

dankolbman:
This removes the dotenv package as we removed the .env from the repo and strictly set everything through the environment in our entrypoint scripts.

It may be worth transitioning to django-environ at a different time.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 import os
 import sys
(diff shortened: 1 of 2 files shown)


XuTheBunny (approved):
🍦

znatty22 (approved):
🆗

gsantia (approved):
📈

PR #534, opened 2021-01-13
🐛 Initialize action should result in pending state

dankolbman:
The spec for release flow says that the `initialize` action sent to a service should result in that service being in the `pending` state, not the `initialized` state.


XuTheBunny (approved):
🍫

PR #536, opened 2021-01-13
🔈 Warn when log file not found

dankolbman:
Prevents a task from failing if an existing log file cannot be found.


XuTheBunny (approved):
🥐

PR #537, opened 2021-01-13
🔈 Make log IO errors less strict

dankolbman:
Catches any exception when trying to read or write logs to ensure that saving logs does not interfere with the wrapped function.


XuTheBunny (approved):
🍑

PR #539, opened 2021-01-15
🔇 Ignore GraphQLError in sentry

dankolbman:
Ignores GraphQLError exception types in sentry as these exceptions are often used to raise intended errors to the user.


XuTheBunny (approved):
🫕

PR #540, opened 2021-01-15
♻️ Don't delete job logs

dankolbman:
Don't delete job logs when the log file cannot be read. Also ensure that deleting job logs won't cascade to releases.


znatty22 (approved):
✅

PR #543, opened 2021-01-15
🔧 Try to remove old job on start up

dankolbman:
Tries to remove the `inventories_sync` job that was never released into production but is still running in dev and qa.


XuTheBunny (approved):
🥟

PR #544, opened 2021-01-15
🐛 Use same user throughout test

dankolbman:
Resolves user by role to ensure the same user is used consistently throughout the test.


znatty22 (approved):
👍

PR #545, opened 2021-01-19
🐛 Handle cases where redis scheduler returns None

dankolbman:
Sometimes redis returns `None` for scheduled tasks which cannot be converted to DateTime. These fields eventually come back, perhaps after `schedule_jobs.py` is run again, so this just handles the odd occurrences where the scheduler doesn't return a value.

### DIFF
--- a/creator/jobs/models.py
+++ b/creator/jobs/models.py
@@ -69,7 +69,12 @@ class Job(models.Model):
         ts = scheduler.connection.zscore(
             "rq:scheduler:scheduled_jobs", self.name
         )
-        dt = datetime.fromtimestamp(ts)
+        try:
+            dt = datetime.fromtimestamp(ts)
+        except TypeError:
+            # Sometimes this is returned as none from redis, perhaps if it is
+            # a one off job
+            return None
         return dt.replace(tzinfo=pytz.UTC)
 
 
--- a/tests/jobs/test_models.py
+++ b/tests/jobs/test_models.py
@@ -1,0 +1,53 @@
+import pytz
+from datetime import datetime
+
+from creator.jobs.models import Job
+
+
+def test_enqueued_at(mocker):
+    """ Test that the enqueued_at property is computed correctly """
+
+    class Scheduler:
+        @property
+        def connection(self):
+            class Connection:
+                def zscore(self, value, name):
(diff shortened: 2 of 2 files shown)


creator/jobs/models.py
```diff
@@ -69,7 +69,12 @@ def enqueued_at(self):
         ts = scheduler.connection.zscore(
             "rq:scheduler:scheduled_jobs", self.name
         )
-        dt = datetime.fromtimestamp(ts)
+        try:
+            dt = datetime.fromtimestamp(ts)
+        except TypeError:
+            # Sometimes this is returned as none from redis, perhaps if it is
```
znatty22, line 75: So what conditions cause redis to return None, do we know?

dankolbman: Redis returns `None` when the key does no exist: https://redis.io/commands/zscore
I need to look into the internals to understand how the scheduler works, but it's also possible that the key is not there because we make the request during the time period between the execution of the task and the scheduling of the next event by the scheduler.

znatty22: So is the purpose of this `enqueued_at` function to get _any_ job's time of insertion into the queue or just scheduled jobs? If its the former, than have you tried querying redis for the job and looking at the enqueued_at property? `Job.fetch('my_job_id', connection=redis).enqueued_at`

dankolbman: The purpose of this property is to get this specific Job's next scheduled queue time. The naming could be better because this is not the same as the `rq.Job().enqueued_at` property. `rq.Job().enqueued_at` specifies the time at which an `rq.Job` was enqueued for run on the _job_ queue. This `enqueued_at` property returns the time on the _scheduler_ queue at which a new `rq.Job()` will be enqueued on the _job_ queue.

znatty22: Ohh I see. Well other than the name of the property this PR looks ok? If you're considering the name maybe: `scheduled_at`

dankolbman: Yeah we could change it to maybe `scheduled_for` or something but not likely in this PR as we have to make changes in the front end, too.

znatty22 (approved):
✅

PR #547, opened 2021-01-21
✉️ Set default from address for tests

dankolbman:
Sets `DEFAULT_FROM_EMAIL` for dev and testing environments so that email functionality will work correctly when using `send_mail()`.


XuTheBunny (approved):
🍨

PR #556, opened 2021-01-28
♻️ Only block new tokens if there is an open invite

### LINKED ISSUE #555: Allow invites to be resent if unclaimed
Invites currently cannot be resent if they have expired as the sending mechanism denies the request if the email has already been sent an invite, regardless of whether it was claimed.
We should check if the existing email has received an invite already and if it has expired, if so, we should regenerate the token and send it in a new email.

dankolbman:
Only returns the 'Waiting for user response' if there is a current _valid_ token in existence for the given email. This allows followup invites to be sent out if the original was allowed to expire.

Closes #555


XuTheBunny (approved):
🍵

PR #558, opened 2021-01-28
✨ Attach users to Sentry

### LINKED ISSUE #557: Identify users for Sentry
It would be helpful if we can add users to Sentry context so we can connect errors with particular users.
See the [docs](https://docs.sentry.io/platforms/python/guides/django/enriching-events/identify-user/) for adding users.

dankolbman:
Attaches user info to Sentry for better event tracking.

Closes #557


znatty22 (approved):
👍

PR #560, opened 2021-01-29
🔊 Rename Sentry transactions with operation name

### LINKED ISSUE #559: Filter /health_check from Sentry transactions
The `/health_check` requests used by ECS should either be completely filtered out or sampled at a reduced rate to avoid sending lots of these transactions to Sentry.

dankolbman:
Splits up the `/graphql` transactions by individual operations to help discern how individual queries are performing.
Ignores `/health_check` transactions.

Closes #559

### DIFF
--- a/creator/settings/production.py
+++ b/creator/settings/production.py
@@ -20,12 +20,25 @@ from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.redis import RedisIntegration
 from graphql import GraphQLError
 
+
+def traces_sampler(sampling_context):
+    """ Filter out unwanted transactions """
+    if (
+        sampling_context.get("wsgi_environ").get("PATH_INFO")
+        == "/health_check"
+    ):
+        return 0
+
+    return 0.1
+
+
 sentry_sdk.init(
     dsn=os.environ.get("SENTRY_URL"),
     environment=os.environ.get("STAGE", "dev"),
     integrations=[DjangoIntegration(), RedisIntegration()],
     ignore_errors=(GraphQLError,),
     traces_sample_rate=0.1,
+    traces_sampler=traces_sampler,
     send_default_pii=False,
 )
 
--- a/creator/settings/production.py
+++ b/creator/settings/production.py
@@ -21,6 +21,9 @@ from sentry_sdk.integrations.redis import RedisIntegration
 from graphql import GraphQLError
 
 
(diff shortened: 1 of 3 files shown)


creator/settings/production.py
```diff
@@ -20,12 +20,25 @@
...
+
+def traces_sampler(sampling_context):
+    """ Filter out unwanted transactions """
+    if (
+        sampling_context.get("wsgi_environ").get("PATH_INFO")
+        == "/health_check"
+    ):
+        return 0
+
+    return 0.1
```
znatty22, line 32: I assume you want to return your default sample rate. So maybe make a constant var for this at the module level and use it here and in `init`

creator/views.py
```diff
@@ -0,0 +1,31 @@
...
+class SentryGraphQLView(FileUploadGraphQLView):
+    """
+    Sets the Sentry transaction name to be that of the GraphQL operation.
+    This helps break down all the /graphql requests into more actionable
+    buckets within Sentry.
+    """
+
+    def execute_graphql_request(self, *args, **kwargs):
+        """
+        Attempt to extract the operation name and use it to set the Sentry
```
znatty22, line 13: Curious, before this change what's an example transaction name and after this what will the transaction names look like?

creator/views.py
```diff
@@ -0,0 +1,31 @@
...
+        # https://github.com/graphql-python/graphene-django/blob/main/graphene_django/views.py#L278-L280
+        if (
+            args
+            and len(args) >= 5
+            and args[4] is not None
+            and len(args[4]) > 0
+        ):
+            from sentry_sdk import Hub
+
+            transaction = Hub.current.scope.transaction
```
znatty22, line 27: When will transaction be None?

znatty22: Also just saw this in the docs. It looks like its doing the same thing but with the context manager: https://docs.sentry.io/platforms/python/guides/django/enriching-events/transaction-name/. Do you want to use that?

dankolbman: The type says it's optional so I'm accounting for that: https://github.com/getsentry/sentry-python/blob/7ba60bda29d671bbef79ae5646fb062c898efc6a/sentry_sdk/scope.py#L146

Yeah I tried the context manager but there was something preventing it from working.

znatty22 (approved):
❄️

PR #562, opened 2021-02-01
🔥 Remove references to the Release Coordinator

dankolbman:
This removes some of the last remaining references to the Release Coordinator. All release functionality is handled natively by the study creator, so we no longer need the Coordinator. The only exception to this is the periodic task that pulls in any releases made inside the old Release Coordinator.


XuTheBunny (approved):
🥟

PR #564, opened 2021-02-04
🐛 Add safeguards for trace sampler

dankolbman:
Sometimes the sampling context won't have everything we need to filter out web requests (like when a task is running). This checks to make sure the right properties exist before filtering.
Also attempts to convert the environment variable for trace sampling rate to a number and will fallback on a hardcoded default if it cannot.


XuTheBunny (approved):
🧁

PR #566, opened 2021-02-09
🐛 Only verify release id if its in the response

dankolbman:
The SBG task service does not respond with a release_id, only a task_id. We should only verify the release_id if it is present.


znatty22 (approved):
✅

znatty22 (approved):
✅

PR #571, opened 2021-02-09
✨ Add new doc types and extract configs

dankolbman:
Document types are defined and described in detail here: 

https://www.notion.so/d3b/Data-Tracker-File-Types-and-Guidelines-1a85b8f9fa1c4469ac44d59d3ccb827e

Adds new document types and extract configs for:
 - Participant Diseases
 - Participant Phenotypes
 - General Observations
 - Biospecimen Collection Manifest
 -  Biobank Manifest
 -  Aliquot Manifest
 - Sequencing File Manifest
 - Complex Family Relationships
 - Family Trios

Updates document types and extract configs:
- Participant Demographics --> Participant Details


creator/extract_configs/templates/participant_observations_config.py
```diff
@@ -0,0 +1,18 @@
+"""
+This is an extract configuration for the Family Trios template
```
znatty22, line 2: Copy paste error?

creator/extract_configs/templates/participant_conditions_config.py
```diff
@@ -0,0 +1,18 @@
+"""
+This is an extract configuration for the Family Trios template
```
znatty22, line 2: Copy paste error?

znatty22 (changes requested):

creator/analyses/file_types.py
```diff
@@ -41,6 +41,28 @@
...
+        "name": "Participant Observations",
+        "required_columns": [
+            "Participant ID",
+            "Age at Onset",
+            "Age at Onset Units",
+            "Observation Name",
+            "Verification Status",
+        ],
+        "template": "participant_observations_config.py",
+    },
```
fiendish, line 65: Per my comment on notion, these conditions and observations forms don't seem to address Kids First needs for diagnosis/phenotype identification.

creator/analyses/file_types.py
```diff
@@ -50,18 +72,27 @@
...
-            'Source Read'
+            "Cavatica ID",
+            "Cavatica Task ID",
+            "KF Biospecimen ID",
+            "KF Participant ID",
+            "KF Family ID",
+            "Filepath",
+            "Data Type",
+            "Workflow Type",
+            "Source Read",
```
fiendish, line 95: Per my comment on notion, most of these fields are useless for ingesting and the useful parts are harder to actually use than necessary (partly a consequence of the mandate to explicitly link SE information from source files to computed files). 

Right now we need to link these GFs to their parent BSes and SEes. Let's imagine a hypothetical future where we either also/instead link to their parent GFs (right now we can ad-hoc use the GF metadata field for that).

If you can get Allison to say that we don't need to make SEGFs for output files (please try because doing it adds a large procedural burden), then we would only need filepath, biospecimen KFIDs (if linking specimens to GFs like we do now), and source files (if linking source GFs to output GFs which we don't do now).

Otherwise the next best option would be to have filepath, source Sequencing Experiment KFIDs, biospecimen KFIDs (if linking specimens to GFs like we do now), and source files (if linking source GFs to output GFs which we don't do now), since those are the things that we have to link.

And if you can't get Sequencing Experiment KFIDs, then you either want biospecimen external IDs instead of KFIDs so that you can put them into the main ingest package and automatically get the SE relationship assignment from the original source data or you have to do an additional round of query processing to discover the SEs from the source GFs before you can ingest the new files.

creator/extract_configs/templates/complex_family_config.py
```diff
@@ -0,0 +1,30 @@
...
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.etl.extract.operations import keep_map
+
+source_data_url = "{{ download_url }}"
+
+
+operations = [
+    keep_map(
+        in_col="Source Participant ID",
+        out_col=CONCEPT.FAMILY_RELATIONSHIP.PERSON1,
```
fiendish, line 20:
```suggestion
        out_col=CONCEPT.FAMILY_RELATIONSHIP.PERSON1.ID,
```

creator/extract_configs/templates/complex_family_config.py
```diff
@@ -0,0 +1,30 @@
...
+    ),
+    keep_map(
+        in_col="Target Participant ID",
+        out_col=CONCEPT.FAMILY_RELATIONSHIP.PERSON2,
```
fiendish, line 24:
```suggestion
        out_col=CONCEPT.FAMILY_RELATIONSHIP.PERSON2.ID,
```

creator/extract_configs/templates/participant_conditions_config.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.etl.extract.operations import keep_map
+
+source_data_url = "{{ download_url }}"
+
+
+operations = [
+    keep_map(in_col="Participant ID", out_col=CONCEPT.PARTICIPANT.ID),
+    keep_map(in_col="Age at Onset", out_col=CONCEPT.DIAGNOSIS.EVENT_AGE_DAYS),
```
fiendish, line 29: You've ignored the age units field.

creator/extract_configs/templates/participant_observations_config.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.etl.extract.operations import keep_map
+
+source_data_url = "{{ download_url }}"
+
+
+operations = [
+    keep_map(in_col="Participant ID", out_col=CONCEPT.PARTICIPANT.ID),
+    keep_map(in_col="Age at Onset", out_col=CONCEPT.DIAGNOSIS.EVENT_AGE_DAYS),
```
fiendish, line 27: You've ignored the age units field and mixed DIAGNOSIS with PHENOTYPE.

creator/extract_configs/templates/participant_observations_config.py
```diff
@@ -0,0 +1,29 @@
...
+    keep_map(in_col="Observation Name", out_col=CONCEPT.PHENOTYPE.NAME),
```
fiendish, line 28: kids first phenotypes must be observed positive or negative.

creator/extract_configs/templates/participant_observations_config.py
```diff
@@ -0,0 +1,29 @@
...
+    Participant ID
+    Age at Onset
+    Age at Onset Units
+    Observation Name
+    Verification Status
+
+Optional Columns:
+    MONDO Code
+    NCIt Code
+    ICD10 Code
```
fiendish, line 16: Is this file nominally for phenotypes? These are not phenotype code systems.

fiendish (changes requested):

creator/extract_configs/templates/participant_conditions_config.py
dankolbman: yeah, is there a way to combine the Age + Units field to convert to `EVENT_AGE_DAYS`?

creator/extract_configs/templates/participant_observations_config.py
dankolbman: Does this mean that we should consider all entries as positive, or require an additional field to annotate it?

fiendish: We have explicitly non-positive entries for "I looked and they definitely don't have X" so either it needs an annotation or we stop doing that.

creator/analyses/file_types.py
dankolbman: These files are generated automatically by a script from the BIX team, so right now they are all marked as required, even it's not the case.

creator/extract_configs/templates/participant_conditions_config.py
fiendish: Yes, the library will handle the conversion automatically if you use CONCEPT.DIAGNOSIS.EVENT_AGE.UNITS and CONCEPT.DIAGNOSIS.EVENT_AGE.VALUE 

See: https://github.com/kids-first/kf-lib-data-ingest/blob/615f935d58eb603b578f326add4b75d153ae8fce/kf_lib_data_ingest/common/misc.py#L18-L31

Unit options are https://github.com/kids-first/kf-lib-data-ingest/blob/615f935d58eb603b578f326add4b75d153ae8fce/kf_lib_data_ingest/common/constants.py#L23-L27

dankolbman: I think we'll need to convert from the FHIR values then? https://www.hl7.org/fhir/valueset-age-units.html#definition
Maybe we can add these extra definitions to the ingest lib units?

fiendish: The codes? Yeah, could do.

creator/analyses/file_types.py
fiendish: I recognize that one purpose of these templates is to handle what we know exists, but another purpose is also to guide what we get. I think this effort should include a conversation with Allison and BIX about what values are most useful.

fiendish:
tests failing. important?

znatty22 (changes requested):
I think this might be a good time to add the ingest lib as a dependency so that we can actually unit test the extract configs for each of these new document types (I can help with that if you want an example or something). For me to implement the validation feature I'm going to need assurance the extract configs work.

Also wondering if we should have a method or utility (e.g. `run_extraction`?), not sure where, (maybe on the File class?) that knows how to load the extract config and run it on the file content to produce the extracted file. This will help us in testing the extract configs and more importantly, we'll need this functionality to produce the dataframes expected by the validation utility in the ingest lib.

znatty22:
Just looking through this a bit more in prep for validation stuff. I noticed that file_types are sort of defined in two places: `creator/analysis/file_types.py` seems to have the full definition and `creator/files/model.py` seems to have just the keys of the former defined in `FileType` Enum. Could `creator/analysis/file_types.py` be moved into `creator/files` and then both the files app and analyses app could import it whenever/wherever needed?

dankolbman:
@znatty22 I agree that we can do some refactoring for the file types, but this PR is to add new file types that we need to comprise a minimal dataset. Perhaps we can do refactor work in a new branch after we've merged this one?

znatty22 (approved):
Approving this based on discussions about implementing tests for the extract configs later. Once https://github.com/kids-first/kf-lib-data-ingest/issues/593 is implemented, testing the extract configs should be very easy. 

Eye-balled the configs and they look good for now!

creator/extract_configs/templates/sample_manifest_config.py
```diff
@@ -0,0 +1,126 @@
...
+    Body Site Code
+    Age at Collection
+    Age at Collection Units
+    Descriptor
+    Analyte Type
+    Method of Sample Procurement
+    Volume
+    Volume Units
+    Concentration
+    Concentration Units
```
fiendish, line 25: Analyte is required not optional.

And what is the plan for when the research group sends a piece of tissue to a sequencing center and then the sequencing center sends part of that tissue to another sequencing center to sequence the same or different analyte or both without assigning new identifiers? This has happened before and will happen again.

creator/extract_configs/templates/sample_manifest_config.py
```diff
@@ -0,0 +1,126 @@
...
+def map_uberon(s):
+    if re.match(UBERON, s):
+        return s
+    return None
+
+
+def map_ncit(s):
+    if re.match(NCIT_RE, s):
+        return s
+    return None
```
fiendish, line 46: same comments as elsewhere re code prefixes, separators, and matching

fiendish (changes requested):

creator/extract_configs/common.py
```diff
@@ -0,0 +1,43 @@
...
+    if re.match(SNOMED_RE, s):
+        return s
+    return None
+
+
+def map_uberon(s):
+    UBERON_RE = r"^UBERON:[0-9]{7}$"
+    if re.match(UBERON_RE, s):
+        return s
+    return None
```
fiendish, line 43: As requested here is some more complex matching that also coerces values to consistent output formats. This code allows truncated input numbers, input prefixes are optional and can be separated by either : or _, and outputs are given consistent prefixes and zero padding.

```suggestion
def _get_match(prefix_pattern, code_pattern, s, result_prefix, result_zeropad=0):
    if prefix_pattern:
        prefix_pattern = f"(?:{prefix_pattern}[:_])?"

    if re.compile(code_pattern).groups == 0:
        code_pattern = f"({code_pattern})"

    capture_pattern = f"^{prefix_pattern}{code_pattern}$"
    assert re.compile(capture_pattern).groups == 1

    match = re.match(capture_pattern, s.strip(), flags=re.IGNORECASE)
    if match:
        return f"{result_prefix}{match.group(1).zfill(result_zeropad)}"
    else:
        return None


def map_mondo(s):
    return _get_match(r"MONDO", r"[0-9]{1,7}", s, "MONDO:", result_zeropad=7)


def map_ncit(s):
    return _get_match(r"NCIT", r"C([0-9]{1,7})", s, "NCIT:C", result_zeropad=7)


def map_icd10(s):
    return _get_match(
        r"ICD(?:10)?", r"[A-TV-Z][0-9][A-Z0-9](?:\.[A-Z0-9]{1,4})?", s, "ICD10:"
    )


def map_hpo(s):
    return _get_match(r"HPO?", r"[0-9]{1,7}", s, "HP:", result_zeropad=7)


def map_snomed(s):
    return _get_match(r"SNOMED", r"[0-9]+", s, "SNOMED:")


def map_uberon(s):
    return _get_match("UBERON", r"[0-9]{1,7}", s, "UBERON:", result_zeropad=7)
```


```Python
def test_matching():
    codes = {
        "MONDO": (
            map_mondo,
            {
                "0000109": "MONDO:0000109",
                "0109": "MONDO:0000109",
                "109": "MONDO:0000109",
                "A": None,
                "12345678": None,
            },
        ),
        "NCIT": (
            map_ncit,
            {
                "C0000100": "NCIT:C0000100",
                "C0100": "NCIT:C0000100",
                "C100": "NCIT:C0000100",
                "1": None,
                "A": None,
                "C12345678": None,
            },
        ),
        "HP": (
            map_hpo,
            {
                "0000100": "HP:0000100",
                "0100": "HP:0000100",
                "100": "HP:0000100",
                "A": None,
                "12345678": None,
            },
        ),
        "HPO": (
            map_hpo,
            {
                "0000100": "HP:0000100",
                "0100": "HP:0000100",
                "100": "HP:0000100",
                "A": None,
                "12345678": None,
            },
        ),
        "ICD": (
            map_icd10,
            {
                "A12": "ICD10:A12",
                "A12.1": "ICD10:A12.1",
                "A13.1114": "ICD10:A13.1114",
                "A": None,
                "12345678": None,
            },
        ),
        "ICD10": (
            map_icd10,
            {
                "A12": "ICD10:A12",
                "A12.1": "ICD10:A12.1",
                "A13.1114": "ICD10:A13.1114",
                "A": None,
                "12345678": None,
            },
        ),
    }
    for prefix, (mapfunc, values) in codes.items():
        for v, res in values.items():
            assert mapfunc(v) == res
            for sep in [":", "_"]:
                assert mapfunc(prefix + sep + v) == res
```

fiendish (changes requested):

creator/extract_configs/templates/family_trio_config.py
```diff
@@ -1,5 +1,13 @@
 """
 This is an extract configuration for the Family Trios template
+
+See: https://www.notion.so/d3b/Expedited-File-Types-and-Guidelines-fc5bd4390fa54de5a70b550d73779de9#9e9e1f8f66184484aa0a187e171365a7
+
+Required Columns:
+    Family ID
+    Mother Participant ID
+    Father Participant ID
+    Proband
```
gsantia, line 10:
```suggestion
    Family ID
    Mother Participant ID
    Father Participant ID
    Proband
    Participant ID
```

creator/extract_configs/templates/participant_config.py
```diff
@@ -1,6 +1,23 @@
...
+
+Required Columns:
+    Participant ID
+    Consent Short Name
+    Affected Status
+    Administrative Gender
+    Ethnicity
+    Age at Study Enrollment Value
+    Age at Study Enrollment Units
+    Deceased
```
gsantia, line 15: Seems like Deceased isn't used in the ingest? Does that make it Optional?

tests/extract_configs/fixtures.py
```diff
@@ -0,0 +1,68 @@
...
+        for req_col in file_type_obj["required_columns"]:
+            if req_col not in df.columns:
+                missing.append(req_col)
+        if missing:
+            raise Exception(
+                f"Test DataFrame for file type {ft} is missing required "
+                f"columns: {missing}. Please modify the DataFrame "
+                "builders in StudyGenerator._df_creators, which can be "
+                "found in creator.studies.data_generator.study_generator"
+            )
```
gsantia, line 56: Do we need to run this part of the test for each different `file_type`? Maybe it could be separate?

creator/extract_configs/templates/participant_conditions_config.py
```diff
@@ -0,0 +1,82 @@
...
+    Age at Onset Value
+    Age at Onset Units
+    Age at Abatement Value
+    Age at Abatement Units
+    Condition Code
+    Body Site Name
+    Body Site Code
+"""
+
+from creator.extract_configs import common
```
dankolbman, line 21: The Study Creator isn't meant to be an installable library and shouldn't be required to run ingest. This config won't work unless the user has study creator in their python path.

I think you want to inline all of the common functions within this template so that user's may use it as an extract config without any requirements other than the ingest library.

dankolbman (commented):
I'm not really sure I understand why the changes to the data generator are necessary. Are these linked to the new document types? When and how should be update these when we make additions of new document types?

creator/extract_configs/templates/participant_observations_config.py
```diff
@@ -0,0 +1,84 @@
+"""
+This is an extract configuration for a Participant Observations file.
```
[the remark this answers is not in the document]
znatty22: cc @allisonheath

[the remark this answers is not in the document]
znatty22: cc @liberaliscomputing

creator/extract_configs/templates/participant_conditions_config.py
znatty22: What do you mean by users? Do we still expect Data Tracker users to be able to download these extract configs and run them in their ingest packages? 

I guess I thought we would only need ingest packages for studies that don't use these new templates. I would think that if the source data files conform to the templates then study creator would run the ingest, in which case, importing the ontology extract functions from outside the config, is ok. If we still expect users to be able to download and use these extract configs in their ingest packages then I understand why we want the common functions inline.

znatty22:
@dankolbman 

> I'm not really sure I understand why the changes to the data generator are necessary. Are these linked to the new document types? 

I need to generate some test data that conforms to the templates so that I can test the extract configs. I do this by running the extract configs on the test data to ensure they run without exception. I decided to modify study generator to create this data. 

I could have done it in a test fixture, but I figured why not add this to study generator? I also updated the study generator's extract configs to make use of the extract configs for the new document types (which I think you requested in a previous PR).  

> When and how should be update these when we make additions of new document types?

The test data should be updated whenever the document templates (`creator/analyses/file_types.py`) are updated so that we can test the companion extract configs

creator/extract_configs/templates/participant_config.py
znatty22: Ah it should be used, I probably just forgot. The required vs optional columns are defined in the file type definitions here: https://www.notion.so/d3b/File-Types-and-Guidelines-fc5bd4390fa54de5a70b550d73779de9

tests/extract_configs/fixtures.py
znatty22: I could move this into another test so it doesn't check every file type every time you use this function

creator/extract_configs/templates/participant_conditions_config.py
znatty22: The ontology extraction functions have been moved to the ingest library so this is no longer an issue

creator/studies/data_generator/ingest_package/extract_configs/complex_family_config.py
```diff
@@ -0,0 +1,12 @@
...
+
+Operations are inherited from the standard Study Creator extract configs in
+creator/extract_configs/templates
+
+The Dataservice entities that will be built from this are:
+    - family_relationship
+"""
+from creator.extract_configs.templates.complex_family_config import operations
+
+source_data_url = "file://../data/bio_manifest.tsv"
```
gsantia, line 12: what's this file for?

gsantia:
Everything looks good to me, but I'm unsure if it's ready for merge? Since the last commit is labeled temp.

znatty22:
Thanks @gsantia. I'm waiting for finalization on the templates' content. I'll have to make changes and then ask for reviews/approvals again.

creator/files/models.py
```diff
@@ -28,17 +32,41 @@ def only_printable(string):
     return "".join([s for s in string if s.isprintable()])
 
 
+def extract_config_path(version):
```
dankolbman, line 35: Where is this used? Why not make it a property of `Version`?

znatty22: Oh you know what this should probably move into the `do-validation` branch where it's actually used. I think I was going to use this function in the tests and then went a different way with tests. I'll move it into `do-validation` and there I can make it part of the `Version` class

znatty22 (approved):
✅

fiendish (changes requested):
My question about why we're asking here for unitless ages + units vs asking for ages in days aside, a short while ago I made a few xlsx documents that could act as the actual template files to be filled in by investigators. We haven't gotten to the stage of finishing the membership rules yet for these let alone the actual labels. I don't think there's any point in merging this PR until we finish creating files like these first.

[Generic_Family_Relationships.xlsx](https://github.com/kids-first/kf-api-study-creator/files/6579931/Generic_Family_Relationships.xlsx)
[Participant_Details.xlsx](https://github.com/kids-first/kf-api-study-creator/files/6579932/Participant_Details.xlsx)
[Trio_Family_Relationships.xlsx](https://github.com/kids-first/kf-api-study-creator/files/6579933/Trio_Family_Relationships.xlsx)

In those files you can see that I changed or expanded column labels, because I think that some of the column headers in this PR obscure their interpretations. They'll all require some amount of clarification either way, but I think that putting more of the explanation into the labels themselves would make them friendlier. Proband and Species are both nouns but one column expects a boolean and one expects a string from a set of given options. "Source participant" and "target participant" don't seem to me to provide a direct enough indication for how to populate "family relationship". Stuff like that.

creator/analyses/file_types.py
```diff
@@ -21,47 +16,188 @@
...
+            "Mother Participant ID",
+            "Father Participant ID"
+        ],
+        "template": "family_trio_config.py"
     },
     "PDA": {
-        "name": "Participant Demographic and Administrative File",
+        "name": "Participant Details",
         "required_columns": [
+            "Family ID",
```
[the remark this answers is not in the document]
znatty22: I think we have to keep this bc Kids First Data Service and the Bix team still use this

creator/analyses/file_types.py
```diff
@@ -21,47 +16,188 @@
...
             "Participant ID",
-            "Consent Short Name",
-            "Administrative Gender",
+            "dbGaP Consent Code",
+            "Clinical Sex",
+            "Gender Identity",
+            "Race",
             "Ethnicity",
-            "Age at Study Enrollment",
+            "Age at Study Enrollment Value",
```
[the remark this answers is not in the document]
znatty22: I remember @allisonheath giving an example of when its useful for age to be in months rather than days..but I can't remember for what field

znatty22:
@fiendish I can change some of the column names per your suggestions (the family relationship changes made sense) but I think we can achieve the clarity for how to populate the field easily with another row for Instructions or something: 

image

creator/analyses/file_types.py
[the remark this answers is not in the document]
fiendish: we could fake them :)

gsantia (approved):
☂

PR #574, opened 2021-02-14
✨ Initial DataReview GraphQL API

### LINKED ISSUE #548: DataReview Model
## Overview
We are adding a new feature to the Data Tracker which allows stakeholders to use Data Tracker to facilitate a Data Review - a process in which stakeholders validate, refine, and ultimately ingest their data into the Kids First Data Service.

A new model should be added to the study creator API to support this feature. This will involve designing the life cycle of the Data Review and the relational model which captures the life cycle and characteristics of the Data Review concept

## Tasks

-   [x] Document describing/illustrating Data Review life cycle
-   [x] Document illustrating Data Review relational model
-   [ ] Implement + test DataReview model in Study Creator

## Design Considerations

-   A Data Review is a process with multiple stages or states (e.g. started, validating, validated, in-review, approved, etc.)

-   A Data Review is related to a set of documents or file versions

-   A Study may have multiple Data Reviews for a unique set of documents before those documents are ingested into production.

1.  What uniquely identifies a Data Review?

    - ❌ A unique set of documents OR a unique set of file versions?
    - ✅ A review will be uniquely identified with an autogenerated unique id. The docs/files will not uniquely identify the review 
since we want the ability to add/remove documents to the same review.

2.  What events/actions cause a Data Review to begin and end?

    - ✅ Begin: After creation of DataReview instance when state goes to started
    - ✅ End : when DataReview state changes to closed or completed

3.  Can a study have multiple "open" Data Reviews at any given point in time?

    - ✅ Yes. Examples: study has multiple batches of documents that we want to review independently

4.  What happens when a document or file version is deleted (or attempted)
    and it is already part of an open DataReview?

        - ❌ Option 1: Not allowed until review is in a closed/completed state
        - ❌ Option 2: Review is automatically closed upon deletion
        - ✅ Option 3: Review stays open, but the related set of documents/files changes

5.  How does a DataReview relate to a "validation run"?

    - A DataReview may have multiple validation runs

6.  What is a "validation run"?

    - ❌ An IngestRun with specific input parameters setup for validation only
    - ✅ A new concept called ValidationRun with a state machine similar to IngestRun. A validation run produces a validation report

7.  What happens when a new file version is added to a document that is part
    of an open Data Review?
- ✅ The user is notified via an event in the timeline

### LINKED ISSUE #549: Data Review GraphQL API
## Overview

Related to #548

We need to add some GraphQL queries and mutations to allow Data Tracker users to create and update DataReviews in Study Creator API. DataReviews should probably never be deleted since they serve as a method of data provenance for the study data.

## Tasks

-   [ ] Add a create mutation to create new DataReviews
-   [ ] Add an update mutation to update fields of the DataReview
-   [ ] Add the usual queries for fetching one DataReview and all DataReviews
-   [ ] Add TBD filter params to the all DataReviews query

znatty22:
Closes #548 
Closes #549 

data_review


creator/data_reviews/signals.py
```diff
@@ -0,0 +1,64 @@
...
+
+@receiver(post_transition, sender=DataReview)
+def data_review_post_transition(
+    sender, instance, name, source, target, *args, **kwargs
+):
+    """
+    Create a corresponding Event after the state of a DataReview is changed.
+    """
+    event = None
+    if (source == State.NOT_STARTED) and (target == State.IN_REVIEW):
```
dankolbman, line 18: Does it make sense to do the event creation inside the model's transitions instead of in a signal?

creator/data_reviews/tasks.py
```diff
@@ -0,0 +1,48 @@
+
+
+def start_review():
```
dankolbman, line 3: I'm not sure about structuring things this way.
Perhaps we instead have tasks for each lengthy action that needs to happen, like emails, and invoke them from the transition or the signal. That way, we don't have all of our tasks executing in one task invocation.

znatty22: Yea I wasn't sure about this either. I can remove the task placeholders

creator/data_reviews/signals.py
znatty22: Hmm, possibly. I guess I can't think of a good reason why the event should be fired _after_ the transition occurs. Although I will have a circular import, since Event imports DataReview. Maybe I can avoid that by putting the Event import inside each transition method?

creator/data_reviews/mutations.py
```diff
@@ -0,0 +1,323 @@
...
+from creator.data_reviews.models import DataReview, State
+
+
+def check_review_files(input, study):
+    """
+    Check if files exist and they are from same study
+    """
+    review_file_ids = []
+    if input.get("files"):
+        study_file_ids = {f.pk for f in study.files.all()}
```
dankolbman, line 18: Maybe premature optimization, but this requires getting all files for the study to validate input files.
You could reduce the amount of data queried by just requesting the kf_ids and comparing those:
```study.files.values("kf_id", flat=True)```
Or you could use the overlap operation:
```
if study.files.filter(kf_id__overlap=input['files']).count() < len(input['files']):
    raise ...
```

File: creator/data_reviews/mutations.py [hunk 4]
```diff
@@ -0,0 +1,323 @@
...
+        """
+        user = info.context.user
+
+        # Check if study exists
+        _, study_id = from_global_id(input["study"])
+        try:
+            study = Study.objects.get(pk=study_id)
+        except Study.DoesNotExist:
+            raise GraphQLError(
+                f"Error in creating data_review. Study {study_id} not found."
```
dankolbman, line 85: I think the 'Error in creating data_review.' is omitted in all our other not found responses.

creator/data_reviews/models.py
```diff
@@ -0,0 +1,152 @@
...
+class DataReview(models.Model):
+    """
+    The iterative process between a data contributor and the Data Resource
+    Center where study data is continually supplied, validated, and refined
+    to meet quality standards.
+    """
+
+    class Meta:
+        permissions = [
+            ("list_all_datareview", "Show all data_reviews"),
```
dankolbman, line 34: do you need to add `add_my_study_datareview`?

creator/data_reviews/queries.py
```diff
@@ -0,0 +1,37 @@
...
+from creator.data_reviews.nodes import DataReviewNode
+from creator.data_reviews.models import DataReview
+
+
+class DataReviewFilter(FilterSet):
+    order_by = OrderingFilter(fields=("created_on",))
+
+    class Meta:
+        model = DataReview
+        fields = []
```
dankolbman, line 15: We probably want to be able to filter on study.

dankolbman (changes requested):

creator/data_reviews/models.py
```diff
@@ -0,0 +1,152 @@
...
+        ]
+
+    kf_id = models.CharField(
```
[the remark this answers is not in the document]
znatty22: I'm leaning yes. I anticipate wanting to send links around (maybe slack) to data reviews

creator/data_reviews/models.py
znatty22: Yea eventually, but right now only the Administrators group can do stuff with data reviews. Which other groups should be able to do stuff for data reviews and which CRUD actions should they be able to do?

creator/data_reviews/models.py
```diff
@@ -0,0 +1,151 @@
...
+    name = models.CharField(
+        max_length=256, help_text="Name of the data review")
+    description = models.TextField(
+        max_length=10000,
+        null=True,
+        blank=True,
+        help_text="Description of data review",
+    )
+    state = FSMField(
+        default="not_started", help_text="The current state of the data review"
```
dankolbman, line 68: Do you want to use `STATE.NOT_STARTED` for the default?

creator/data_reviews/models.py
```diff
@@ -0,0 +1,151 @@
...
+from creator.files.models import File
+
+User = get_user_model()
+
+
+def data_review_id():
+    return kf_id_generator("DR")
+
+
+class State(object):
```
dankolbman, line 17: Would this also work as an Enum?

creator/data_reviews/mutations.py
```diff
@@ -0,0 +1,323 @@
...
+                if attr in input:
+                    setattr(data_review, attr, input[attr])
+
+            if review_file_ids:
+                # Start review if not started
+                if data_review.state == State.NOT_STARTED:
+                    data_review.start()
+
+                # We received updates while waiting, go back to reviewing state
+                elif data_review.state == State.WAITING:
```
dankolbman, line 181: So if we update the review's name or description while it's in the `waiting` state, it will go back to the `reviewing` state even if the files didn't change? I think we need to check if the new list of files is different from the old list of files in order to make this transition.

creator/data_reviews/models.py
```diff
@@ -0,0 +1,151 @@
...
+    state = FSMField(
+        default="not_started", help_text="The current state of the data review"
+    )
+    study = models.ForeignKey(
+        Study,
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="data_reviews",
+    )
+    files = models.ManyToManyField(
```
dankolbman, line 76: I think we want to be relating to versions here? Having files here would imply we are always using the latest version for the review, which may not be true. Furthermore, once a review is completed, uploading a new version of one of the files would sorta 'corrupt' the review.

dankolbman (changes requested):

creator/data_reviews/models.py
znatty22: Ah oops yea

creator/data_reviews/mutations.py
znatty22: If the name/description change it has no effect on the state of the review bc this code only executes if "files" are in the input. However, you bring up a good point. I should only update from waiting to in review if the files actually changed. Right now that state is just based on the presence of "files" in the input and whether they're all valid.

creator/data_reviews/models.py
znatty22: Yes, but then I have to type out for example, `State.NOT_STARTED.value`, every time I want to access the value of the enumerated item

creator/data_reviews/mutations.py
dankolbman: What if we have a review with state `{"name": "ABC", "files": ["A","B"]}`.
Then we receive an update with payload `{"name": "123", "files": ["A","B"]}`.
Wouldn't this cause it to back to the reviewing state despite the files not changing?

znatty22: No, bc I don't think name/description changes should trigger the data review state to change

creator/data_reviews/signals.py
```diff
@@ -0,0 +1,98 @@
...
+    Check if instance.root_file has a version that is part of any Data Review
+    Yes - update all related review's state, emit event
+    """
+    # Only process new versions with root files
+    if (not created) or (not instance.root_file):
+        return
+
+    # For data reviews involving the root file of the version that was uploaded
+    # update the reviews' state + emit events
+    for dr in DataReview.objects.filter(
```
dankolbman, line 80: Why can't you do the following here?
```
for dr in instance.data_reviews.filter(state=State.WAITING).all():
```

creator/data_reviews/signals.py
```diff
@@ -0,0 +1,98 @@
...
+        versions__in=Version.objects.filter(root_file=instance.root_file)
+    ).distinct():
+        if dr.state == State.WAITING:
+            dr.receive_updates()
+            dr.save()
+        Event(
```
dankolbman, line 86: I think you only need to emit this event if the data review state changes? Otherwise, no change is being made to the data review, so no need to make an event.

dankolbman (changes requested):

creator/data_reviews/signals.py
znatty22: My thought was even if the dr state doesn't change, we may still want to emit an event to alert the user that a new version was uploaded, if that new version is linked to a file in the review. I imagine a couple scenarios. The first is the obvious one where feedback has been sent to the investigator and we're waiting for updates. When we receive the updates we definitely want to make it known to the user that updates have been received. But then another scenario is, maybe the review is still IN_REVIEW and an analyst makes an update to a file (perhaps its something easy they can do without input from investigator, etc). I think we still want to emit this event to make it obvious that a change has been made to a file in the review.  

Perhaps what we could do is emit the event only if the review is not in closed or completed state? However, I would ask - do we actually still want to emit this event even if the review is closed/completed? I can imagine still getting updates to files even after the review is closed or completed and wanting some sort of record of this?

creator/data_reviews/signals.py
znatty22: See comment above

creator/data_reviews/mutations.py
```diff
@@ -0,0 +1,354 @@
...
+    if len(studies) > 1:
+        raise GraphQLError(
+            "Error in modifying data_review. All file versions in data "
+            "review must have files that belong to the same study."
+        )
+
+    # Check if data review file versions changed
+    if data_review:
+        intersect = len(
+            data_review.versions.filter(pk__in=input_version_ids).all()
```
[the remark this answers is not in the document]
znatty22: Ah oops, yes

creator/data_reviews/signals.py
znatty22: Thinking some more..I have to go back and look at what that Data Review detail view will look like. If the event timeline will already have events about new versions being uploaded (like the File detail page) then maybe this event is redundant?

dankolbman: Ok I think I understand the logic here better now. I would say this is good, except for you probably want to only emit events for data reviews that are not in terminal states. It would be strange to see a Data Review marked as approved or closed and then have events following that.

dankolbman (approved):
# 👺

PR #579, opened 2021-02-17
🐛 Fix bugs from switching to Debian image

znatty22:
Tested locally with docker-compose on the ingest-app branch. Ran the start ingest run mutation and the worker looks like it picks up the job and executes successfully.


dankolbman (approved):
# 🥭

dankolbman (approved):
# 👍

PR #581, opened 2021-02-23
✨ Add investigator name to study

### LINKED ISSUE #78: Add Investigator/StudyContact Node
Each study should connect to a some sort of owner/user/investigator (I'll leave the naming up to you) node. This will allow us to group by and manage studies by investigator. 

Use case query: 
```js 
{
    allStudies {
      edges {
        node {
          id
          kfId
          name
          investigator
        }
      }
    }
  }
```

dankolbman:
Adds an investigator name field to the study model so we can track the name of the investigator.

Closes #78


znatty22 (commented):
How does this relate to what's in dataservice? Will we eventually use this to populate the study investigators in dataservice?

dankolbman:
For Data Service, I think we will ignore the investigator model that we have there. We are tending towards having the Study Creator/Data Tracker be the authoritative source for studies as that is where studies are created.

znatty22:
Would you want to consider syncing investigators from the dataservice initially since a lot of the studies have the investigators filled out there? You could add some logic to the `sync_dataservice_studies` method that pulls in all the investigators for the studies and then populates the creator's Study.investigator_name if it is null?

znatty22 (approved):
👨🏼‍🔬

PR #592, opened 2021-03-02
✨ Set download endpoints to use http in dev mode

### LINKED ISSUE #585: Download endpoints should use http in dev mode
If developers are running locally, they should be able to test and use the `/download` endpoint. This currently requires code modification because all download urls are returned with the https:// protocol. This is required to be hardcoded because our production environment does not terminate SSL itself.

We should check if `DEVELOP = True` in our settings and then return `http://` instead of `https://`.

Here are the locations where the protocol is currently hardcoded:
https://github.com/kids-first/kf-api-study-creator/blob/8dbae2a0691f978bf5c22b9b22ac7e064ca9c43d/creator/files/nodes/version.py#L21

https://github.com/kids-first/kf-api-study-creator/blob/6e000e7b2f82a318e25cf549b4701ecfc45d591d/creator/files/nodes/file.py#L38

https://github.com/kids-first/kf-api-study-creator/blob/6d7743135dd5cee2a4bad57397f043e4fed8d006/creator/jobs/schema.py#L64

gsantia:
Allow developers running the study creator locally to test and use the `/download` endpoint. This is done by checking
if the `DEVELOP` setting is `True`. If so, return `http://` instead of `https://` in the download URLs.

Closes #585.

### DIFF
--- a/creator/files/nodes/file.py
+++ b/creator/files/nodes/file.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 import graphene
 from graphene import relay
 from graphene_django import DjangoObjectType
@@ -35,7 +36,8 @@ class FileNode(DjangoObjectType):
     valid_types = graphene.List(FileType)
 
     def resolve_download_url(self, info):
-        return f"https://{info.context.get_host()}{self.path}"
+        protocol = "http" if settings.DEVELOP else "https"
+        return f"{protocol}://{info.context.get_host()}{self.path}"
 
     @classmethod
     def get_node(cls, info, kf_id):
--- a/creator/jobs/schema.py
+++ b/creator/jobs/schema.py
@@ -4,6 +4,7 @@ from graphene import relay
 from graphene_django import DjangoObjectType
 from graphene_django.filter import DjangoFilterConnectionField
 from django_filters import FilterSet, OrderingFilter
+from django.conf import settings
 from graphql import GraphQLError
 
 from creator.jobs.models import Job, JobLog
@@ -61,7 +62,8 @@ class JobLogNode(DjangoObjectType):
     download_url = graphene.String()
 
     def resolve_download_url(self, info):
-        return f"https://{info.context.get_host()}{self.path}"
+        protocol = "http" if settings.DEVELOP else "https"
+        return f"{protocol}://{info.context.get_host()}{self.path}"
 
     @classmethod
     def get_node(cls, info, id):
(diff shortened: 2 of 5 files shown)


creator/files/nodes/file.py
```diff
@@ -35,7 +36,8 @@ class Meta:
     valid_types = graphene.List(FileType)
 
     def resolve_download_url(self, info):
-        return f"https://{info.context.get_host()}{self.path}"
+        HTTP = "http" if settings.DEVELOP else "https"
```
dankolbman, line 39: Perhaps a better name for this variable would be `protocol`?

creator/jobs/schema.py
```diff
@@ -1,9 +1,10 @@
 import django_filters
+from django_filters import FilterSet, OrderingFilter
```
dankolbman, line 2: Do you need to re-order these imports here?

docker-compose.dataservice.yml
```diff
@@ -15,6 +15,8 @@ services:
       - PG_HOST=dataservice-pg
     depends_on:
       - dataservice-pg
+    ports:
```
dankolbman, line 18: I don't think this needs to change in this PR?

dankolbman:
Can you add a description to the PR?

creator/jobs/schema.py
gsantia: nope. i like having them order alphabetically

docker-compose.dataservice.yml
gsantia: Sorry, some extra files got added in. Removed

tests/files/test_downloads.py
```diff
@@ -78,6 +79,41 @@ def test_file_download_url(clients, db, mocker):
...
+    )
+    file_json = resp.json()["data"]["allFiles"]["edges"][0]["node"]
+    expect_url = (
+        f"http://testserver/download/study/{study.kf_id}/file/{file.kf_id}"
+    )
+    assert resp.status_code == 200
+    assert "data" in resp.json()
+    assert "allFiles" in resp.json()["data"]
+    assert file_json["downloadUrl"] == expect_url
+    resp = client.get(file_json["downloadUrl"])
```
dankolbman, line 109: You can probably remove all of the bits about downloading the actual file. This is tested in other tests above.

dankolbman (changes requested):
Can you add another test to test the job log downoads? It should be pretty similar.

dankolbman (changes requested):
# 🎢

Nice! Been looking forward to this so I don't need to modify the urls whenever I'm testing locally.

dankolbman (approved):
# 🎢

Nice! Been looking forward to this so I don't need to modify the urls whenever I'm testing locally.

PR #595, opened 2021-03-03
👷 Add new workflow for sentry releases

dankolbman:
Makes the Sentry action only run on tagged release commits and ignores issues when no commits are found.

### DIFF
--- a/.github/workflows/gh_release.yml
+++ b/.github/workflows/gh_release.yml
@@ -56,14 +56,6 @@ jobs:
                 "${upload_url%\{*}?name=$(basename $FILE)"
             done < .github/release_assets.txt
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/gh_release.yml
```diff
@@ -57,13 +57,6 @@ jobs:
             done < .github/release_assets.txt
           fi
       - name: Create Sentry release
-        uses: getsentry/action-release@v1
```
znatty22, line 60: Does line 59 (`- name: Create Sentry release`) have to be removed too?

gsantia:
Sorry I'm going to need to learn about Github Actions before I can comment 😬

znatty22 (approved):
📦

PR #599, opened 2021-03-11
Review + suggestions for ingest-task-genomic-workflow-output

znatty22:
TODO
- Fix migration issue
- Create a study in all three deployed envs for ingest
- Update data files with correct study id from ^
- Provided suggestions/review for unit tests
- Add a sanity check / assertions to check that ingest_gwo worked correctly


creator/ingest_runs/data_generator/ingest_package/extract_configs/biospec.py
```diff
@@ -22,14 +22,20 @@
 
 # TODO - Replace this with operations that make sense for your own data file
 operations = [
-    value_map(
```
znatty22, line 25: We shouldn't mutate the original id so just use keep map for all the external ids coming from our generated study

creator/ingest_runs/data_generator/ingest_package/extract_configs/genomic.py
```diff
@@ -5,24 +5,20 @@
...
-import pandas as pd
-
-from kf_lib_data_ingest.common import constants  # noqa F401
+import os
+from kf_lib_data_ingest.common import constants, pandas_utils  # noqa F401
 from kf_lib_data_ingest.common.concept_schema import CONCEPT
 from kf_lib_data_ingest.etl.extract.operations import *
+from kf_lib_data_ingest.common.io import read_df
 
-# TODO - Replace this with a URL to your own data file
```
znatty22, line 14: Instead of incorporating the file metadata this way, we can do this separately in another extract config. We just use the standard s3 scrape extract config to extract the s3 file metadata and then merge it into our df in the transform module.

creator/ingest_runs/data_generator/ingest_package/extract_configs/genomic.py
```diff
@@ -33,46 +29,25 @@ def do_after_read(df):
 
 # TODO - Replace this with operations that make sense for your own data file
 operations = [
-    value_map(
+    keep_map(
```
znatty22, line 32: Same comment as above about mutating external ids

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
+"""
```
znatty22, line 1: I took all of the individual methods you had and put them in one class so its easier to see what's going on and call in an entry point function that will do the data generation and the ingestion of the files into Data Service.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+from kf_lib_data_ingest.etl.ingest_pipeline import DataIngestPipeline
+from d3b_utils.aws_bucket_contents import fetch_aws_bucket_obj_info
+
+# For data generation
+ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+DATA_DIR = os.path.join(ROOT_DIR, "data_generator", "ingest_package", "data")
+INGEST_PKG_DIR = os.path.join(ROOT_DIR, "data_generator", "ingest_package")
+
+# Connections
+DATASERVICE_URL = "http://localhost:5000"
```
znatty22, line 38: We will eventually need to get this from the django app settings and not hard code it here.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+            "biospecimen": "BS",
+            "genomic_file": "GF",
+        }
+
+    @property
+    def study_bucket(self):
+        """ Return study bucket from study KF ID """
+
+        return TEMP_STUDY_BUCKET
+        # return f"kf-study-us-east-1-prd-{study_id.lower().replace('_', '-')}"
```
znatty22, line 86: We need to make a study bucket in all the dev, qa, prd deployed environments for ingest testing, ingest the data files into that bucket, and then we can replace this with the commented line below.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+        ra.seed(0)
+        self.logger.info(
+            f"Creating clinical and genomic files for {self.study_id}"
+        )
+
+        # Create clinical and genomic data files
+        bio_df = self._create_bio_manifest(create=create)
+        gf_df = self._create_gfs(create=create)
+
+        # Upload unharmonized and harmonized gfs to S3
```
znatty22, line 130: I added an upload method so we don't have to manually upload the files. It was getting tedious so decided to do it via code after all.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+            self.logger.info(f"Creating {url}")
+            try:
+                resp = requests.post(url, json=p)
+                resp.raise_for_status()
+            except requests.exceptions.HTTPError:
+                ok = resp.status_code == 400 and "already exists" in resp.text
+                if not ok:
+                    raise
+
+    def _kf_id_generator(self, prefix):
```
znatty22, line 189: This should probably be imported from `creator/fields.py` oops

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+        bio_dict = {
+            "sample_id": [f"SM-{i}" for i in _range],
+            "participant_id": [f"CARE-{i}" for i in _range],
+            "gender": [ra.choice(("Female", "Male")) for _ in _range],
+            "volume": [100] * n,
+            "concentration": [30] * n,
+            "family_id": [ra.choice(("FA-1", "FA-2")) for _ in _range],
+            "tissue_type": [ra.choice(("Blood", "Saliva")) for _ in _range],
+        }
+        id_gen = self._kf_id_generator
```
znatty22, line 236: This method is the same as before except that I added the logic to inject the fixed kf ids

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+
+        os.makedirs(self.data_dir, exist_ok=True)
+        df = pd.DataFrame(bio_dict)
+        df.to_csv(fp, sep='\t', index=False)
+
+        self.logger.info(f"Wrote bio manifest to {fp}")
+
+        return df
+
+    def _create_gfs(self, create=False):
```
znatty22, line 256: I forget what the original methods were, but this method essentially combines the creation of the genomic files on disk and the writing of a genomic file manifest. It just seemed easier to create the manifest as the files are being written.

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+                )
+                rows.append(row_dict)
+
+        df = pd.DataFrame(rows)
+        df.to_csv(manifest_fp, sep='\t', index=False)
+        self.logger.info(f"Wrote {manifest_fp}")
+
+        return df
+
+    def _upload_gfs_to_s3(self):
```
znatty22, line 347: New method

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+        s3_data = fetch_aws_bucket_obj_info(
+            bucket_name=self.study_bucket,
+            search_prefixes=[
+                "source/genomic-files/",
+                "harmonized/genomic-files/"
+            ],
+        )
+        df = pd.DataFrame(s3_data)
+
+        # Filter out folders
```
znatty22, line 399: Had to do this since they removed that `drop_folders` kwarg from the fetch function

creator/ingest_runs/data_generator/study_generator.py
```diff
@@ -0,0 +1,504 @@
...
+            list(col_map.keys()) +
+            ["Data Type", "Workflow Type", "Cavatica ID", "Cavatica Task ID"]
+        ].rename(columns=col_map)
+
+        df.to_csv(fp, sep="\t", index=False)
+        self.logger.info(f"Wrote {fp}")
+
+        return df
+
+    def _assert_load(self, endpoint, kf_ids):
```
znatty22, line 473: Added a sanity check to make sure that all of the KF IDs we generated in the files were actually loaded into the Data Service

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
+import ast
```
znatty22, line 1: I renamed `creator/ingest_runs/manifests.py` to `creator/ingest_runs/genomic_data_loader.py` bc it seems more descriptive/intuitive.

Also wrapped most of the methods from the manifest module in a class. Class is useful so we can save access common objects more easily like the logger, the loader, etc.

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+import pandas as pd
+
+from kf_lib_data_ingest.app import settings as ingest_settings
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.common.constants import GENOMIC_FILE
+from kf_lib_data_ingest.etl.load.load import LoadStage
+
+from creator.ingest_runs import utils
+
+FILE_EXT_FORMAT_MAP = {
```
znatty22, line 15: This should probably be put in a common module that this module and the s3_scrape_config.py can both access. Write now the same structure is defined in two diff places. Could maybe go in creator/ingest_runs/utils.py?

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+from pprint import pprint
+
+import pandas as pd
+
+from kf_lib_data_ingest.app import settings as ingest_settings
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.common.constants import GENOMIC_FILE
+from kf_lib_data_ingest.etl.load.load import LoadStage
+
+from creator.ingest_runs import utils
```
znatty22, line 13: Separated out some utility functions from the genomic data loading

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+            f"{DATASERVICE_URL}"
+        )
+
+        # Load harmonized genomic files
+        genomic_cache = self._load_entities(GEN_FILE, df)
+        genomic_df = self._get_genomic_file_kf_ids(genomic_cache, genomic_df)
+
+        return genomic_df
+
+    def load_specimen_harmonized_gf_links(self, genomic_df):
```
znatty22, line 145: Decided to put this in its own function since the other entities had their own load functions

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+SEQ_EXP = "sequencing_experiment"
+SEQ_EXPS = "sequencing-experiments"
+SEQ_EXP_GEN_FILE = "sequencing_experiment_genomic_file"
+SEQ_EXP_GEN_FILES = "sequencing-experiment-genomic-files"
+BIO_GEN_FILE = "biospecimen_genomic_file"
+BIO_GEN_FILES = "biospecimen-genomic-files"
+LOAD_ENTITY_TYPES = [GEN_FILE, BIO_GEN_FILE, SEQ_EXP_GEN_FILE]
+
+LOCALHOST = "http://localhost:5000"  # testing
+DATASERVICE_URL = LOCALHOST  # During testing
```
znatty22, line 46: Same comment as before. This should come from the django app settings

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+        )
+        self.logger.info(
+            f"Creating {df.shape[0]} sequencing-experiment-genomic-file "
+            f"links in {DATASERVICE_URL}"
+        )
+        _ = self._load_entities(SEQ_EXP_GEN_FILE, df)
+
+        return se_gf_df
+
+    def _prep_harmonized_gf_df(self, df):
```
znatty22, line 213: Renamed munge_genomic_df to this

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+        """
+        Part of Step 1
+
+        Clean up the genomic file DataFrame (output from get_s3_file_metadata)
+        and get it ready to be ingested by transforming column values and
+        standardizing the column names.
+        """
+        self.logger.info("Preparing genomic file DataFrame for ingest ...")
+
+        def file_ext(x):
```
znatty22, line 223: Added this to extract the file extension/format. This is what the s3_scrape_config.py extract config uses. We should probably put this in a common module and then import it here rather than define it again

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+            ],
+        )
+        genomic_df = genomic_df.merge(
+            right=hash_df,
+            on=CONCEPT.GENOMIC_FILE.ID,
+            how='inner',
+        )
+        return genomic_df
+
+    def _get_seq_experiment_genomic_files(self, harmonized=False):
```
znatty22, line 290: Simplified this function (and renamed it) because the previous version was doing things with biospecimens and I don't think we need any biospecimen info for this

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,397 @@
...
+            self.target_api_cfg,
+            DATASERVICE_URL,
+            [entity_type],
+            self.study_id,
+            cache_dir=os.getcwd(),
+        )
+        self.loader.run({entity_type: df})
+        return self.loader.uid_cache
+
+    def _validate_load(self, df):
```
znatty22, line 395: We should have a sanity check that will be called after the whole ingest process. Don't know what should go in here yet

creator/ingest_runs/tasks.py
```diff
@@ -55,23 +55,28 @@ def check_gwo(version):
...
+    """
+    TODO - docstring
+    """
     ingest_run = IngestRun.objects.get(pk=ingest_run_uuid)
     logging.info(f"Canceling ingest run {ingest_run.pk}...")
     ingest_run.cancel()
     ingest_run.save()
 
 
 def ingest_genomic_workflow_output_manifests(ingest_run):
```
znatty22, line 67: Filled in the TBD parts to extract file content from the file version object into a DataFrame

creator/ingest_runs/data_generator/study_generator.py
gsantia: for some reason, I couldn't get this function to import, so I just copied it over

creator/ingest_runs/manifests.py
```diff
@@ -1,334 +0,0 @@
...
-    assert "Filepath" in manifest_df.columns, "Missing Filepath column"
-
-    buckets = set(os.path.split(row)[0] for row in manifest_df["Filepath"])
-    file_df = scrape_s3(buckets)
-
-    genomic_df = manifest_df.merge(right=file_df, on="Filepath", how="inner")
-    return genomic_df
-
-
-def scrape_s3(buckets):
```
znatty22, line 80: Moved this into a new module creator.ingest_runs.utils

creator/ingest_runs/manifests.py
```diff
@@ -1,334 +0,0 @@
...
-    """
-    genomic_df = munge_genomic_df(genomic_df)
-    genomic_cache = load_entities(
-        GEN_FILE,
-        genomic_df.drop(CONCEPT.BIOSPECIMEN.ID, axis=1),
-        study_id,
-    )
-    genomic_df = get_genomic_file_kf_ids(genomic_cache, genomic_df)
-    
-    biospec_gen_df = genomic_df[[
```
znatty22, line 121: I liked that you seemed to have a load_<entity type>_  method for each type, so I pulled out the loading of the biospecimen-genomic-files into its own load method too

creator/ingest_runs/manifests.py
```diff
@@ -1,334 +0,0 @@
...
-
-
-def munge_genomic_df(df):
-    """
-    Clean up the genomic-file DataFrame and get it ready to be ingested by
-    removing unneeded columns and standardizing the column names.
-    """
-    df['Hashes'] = df.apply(lambda row: {"ETag": row['ETag']}, axis=1)
-    df['urls'] = df.apply(lambda row: [row['Filepath']], axis=1)
-    df['file_type'] = df.apply(
```
znatty22, line 185: This would have been hard for you to know without doing a bunch of ingests, but getting the file extension for the genomic files is a little tricky. There is a function in the s3_scrape_config.py which does this, so I replaced your lambda with that function. See _prep_harmonized_gf_df in creator/ingest_runs/genomic_data_loader.py

creator/ingest_runs/manifests.py
```diff
@@ -1,334 +0,0 @@
...
-    for entity_type in entity_types:
-        dfs[entity_type] = get_entities(entity_type, study_id)
-
-    # Step 5. Join tables 1-4 on genomic file ID. Some preprocessing needs to
-    # be done first.
-    
-    # Start by merging the DataFrames from steps 1 and 3 on genomic-file ID
-    GEN_LINK = '_links.genomic_file'
-
-    # Check the validity of the SEQ_EXP_GEN_FILES DataFrame
```
znatty22, line 234: I removed these assertions because I think it's safe to assume the output of the kf_utils.yield_entities method won't change.

creator/ingest_runs/manifests.py
```diff
@@ -1,334 +0,0 @@
...
-    assert (
-        GEN_LINK in dfs[SEQ_EXP_GEN_FILES], missing_data(SEQ_EXP_GEN_FILES)
-    )
-    #TODO: Do we want to check here that all values for genomic-file ID are
```
znatty22, line 238: I want to understand what you meant by this. I think in general, we should talk about what kinds of error checking/validation we want to do. I'm thinking wrap each of the top level `load_<entity type>` functions in a try except. In the except block we can log some details and then reraise the exception. Not sure tho, let's talk about this

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -0,0 +1,361 @@
...
+        sedf = dfs[SEQ_EXPS][
+            ["kf_id", "external_id", "_links.sequencing_center"]
+        ].rename(
+            columns={
+                "kf_id": CONCEPT.SEQUENCING.TARGET_SERVICE_ID,
+                "external_id": CONCEPT.SEQUENCING.ID,
+            }
+        )
+        SCTID = CONCEPT.SEQUENCING.CENTER.TARGET_SERVICE_ID
+        sedf[SCTID] = sedf["_links.sequencing_center"].apply(
```
znatty22, line 313: Gio says use .map over .apply for single col transformations bc its 10-100x faster

PR #608, opened 2021-03-25
✅ Tests for ingest-app branch and ingest GWO feature flag

gsantia:
Increase code coverage on the ingest-app branch so that we can merge it into master. Add a `FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS` feature flag and associated test which allows for the ingestion of GWO to be toggled on/off.

### DIFF
--- a/creator/ingest_runs/tasks.py
+++ b/creator/ingest_runs/tasks.py
@@ -1,6 +1,7 @@
 import django_rq
 import logging
 
+from django.conf import settings
 from django.core.exceptions import ValidationError
 
 from creator.decorators import task
@@ -29,6 +30,16 @@ def run_ingest(ingest_run_uuid=None):
     versions = ingest_run.versions.all()
     are_gwo = all(check_gwo(version) for version in versions)
     if are_gwo:
+        # If FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS is enabled, try to run
+        # the custom ingest process for genomic workflow manifests. If it's not
+        # enabled, raise an error and nothing else happens.
+        if not settings.FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS:
+            ingest_run.fail()
+            ingest_run.save()
+            raise Exception(
+                "Ingesting genomic workflow output manifests is not enabled."
+                "Make sure that FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS is set."
+            )
         logging.info(
             "All files in the ingest run are genomic workflow manifests."
         )
--- a/creator/ingest_runs/tasks.py
+++ b/creator/ingest_runs/tasks.py
@@ -1,13 +1,10 @@
-import django_rq
 import logging
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
 
 from creator.decorators import task
 from creator.files.models import FileType
 from creator.ingest_runs.models import IngestRun
-from creator.utils import stop_job
 
 
 logger = logging.getLogger(__name__)
@@ -71,17 +68,4 @@ def cancel_ingest(ingest_run_uuid=None):
 
 
 def ingest_genomic_workflow_output_manifests(ingest_run):
-    # TODO
-    print(
-        f"Begin ingesting genomic workflow manifests: "
-        f"{list(ingest_run.versions.all())}"
-    )
-    import time
-
-    t_end = time.time() + 10
-    progress = ""
-    while time.time() < t_end:
-        time.sleep(1)
-        progress += "."
-        print(progress)
-    print(f"Finished ingest run {ingest_run.id}")
+    pass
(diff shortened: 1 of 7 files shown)


znatty22:
Hey @gsantia could you rename this PR and update the description to include the feature flag

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -208,3 +229,52 @@ def test_cancel_ingest_run(
...
+    ir1.versions.set(file_versions)
+    ir1.save()
+    ir2 = IngestRun()
+    ir2.save()
+    ir2.versions.set(file_versions)
+    ir2.save()
+    irs = (ir1, ir2)
+    assert IngestRun.objects.all().count() == 2
+    cancel_duplicate_ingest_runs(irs)
+    assert mock_ingest_queue.call_count == 2
```
znatty22, line 280: Maybe add a check that both ingest runs have a state of canceled?

tests/ingest_runs/test_tasks.py
```diff
@@ -82,6 +82,57 @@ def test_run_ingest(db, mocker, clients, prep_file):
...
+    settings.FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS = False
+    mock_genomic_workflow = mocker.patch(
+        "creator.ingest_runs.tasks.ingest_genomic_workflow_output_manifests"
+    )
+
+    user = User.objects.first()
+    # Create data.
+    for _ in range(3):
+        prep_file(authed=True)
+    files = list(File.objects.all())
```
znatty22, line 100:
```suggestion
    versions = []
```

tests/ingest_runs/test_tasks.py
```diff
@@ -82,6 +82,57 @@ def test_run_ingest(db, mocker, clients, prep_file):
...
+    for file_ in files:
+        file_.file_type = FileType.GWO.value
+        file_.save()
```
znatty22, line 103:
```suggestion
        file_.save()
        versions.append(file.versions.first())
```

tests/ingest_runs/test_tasks.py
```diff
@@ -82,6 +82,57 @@ def test_run_ingest(db, mocker, clients, prep_file):
...
+    versions = [f.versions.first() for f in files]
```
znatty22, line 104:
```suggestion

```

znatty22 (changes requested):

tests/ingest_runs/test_ingest_runs_mutations.py
gsantia: They won't, because we're using `mock_ingest_queue`

znatty22: oh I see

creator/ingest_runs/tasks.py
```diff
@@ -29,6 +29,16 @@ def run_ingest(ingest_run_uuid=None):
     versions = ingest_run.versions.all()
     are_gwo = all(check_gwo(version) for version in versions)
     if are_gwo:
+        # If FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS is enabled, try to run
+        # the custom ingest process for genomic workflow manifests. If it's not
+        # enabled, raise an error and nothing else happens.
+        if not settings.FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS:
```
dankolbman, line 35: An alternative approach would be to check the flag immediately inside the task and then bailout then or even check the flag before invoking the task. If you check the flag during the mutation, then you can avoid queuing up the task all together.

znatty22: @dankolbman We're checking it here because its specifically to enable/disable the GWO ingest not all ingests. I suppose we could have a specific mutation like StartGWOIngest that queues a run_gwo_ingest task and then we could use your approach. But then that's some more implementation and refactoring ... might not be worth it now / better to wait until we implement the rest of ingest.

znatty22 (approved):
☀️

creator/settings/features.py
```diff
@@ -158,6 +158,15 @@
 )
 
 
+# GWO INGEST RUNS ##############################################################
+# The Study Creator can automate various ingest processes such as ingesting
+# harmonized genomic files.
+
+FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS = os.environ.get(
+    'FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS', 'True'
```
dankolbman, line 166: I think black should format this with double quotes?

gsantia: that's true. what black options should i use in general? i notice that when you just run `black` it sometimes makes lines too long

dankolbman: Black should always format consistently except for those weird cases where it won't adjust the line length for you. In those situations I usually have to break up the line manually and then let black format it nicely.

dankolbman (approved):
# 🍡

PR #609, opened 2021-03-25
✨ Validation models

### LINKED ISSUE #550: Validation Models

znatty22:
Adds models, events, and signals to capture the process of validating study files and reporting results.

Closes #550 

<p align="center">
image
</p>
<p align="center">
image
</p>


creator/ingest_runs/common/model.py
```diff
@@ -0,0 +1,182 @@
...
+    created_at = models.DateTimeField(
+        auto_now_add=True,
+        null=True,
+        help_text="Time when the ingest process was created",
+    )
+    creator = models.ForeignKey(
+        User,
+        null=True,
+        blank=True,
+        related_name="%(class)ss",
```
gsantia, line 63: I'm a bit confused here, why no `'%(app_label)s'`?

creator/ingest_runs/common/model.py
```diff
@@ -0,0 +1,182 @@
...
+        running for a set of file versions
+        """
+        return hash_versions(self.versions.all())
+
+    def _save_event(self, event_type):
+        """ Create and save an event for an ingest process state transition """
+        from creator.events.models import Event
+        snake_name = camel_to_snake(self.__class__.__name__)
+        name = camel_to_snake(self.__class__.__name__).replace("_", " ")
+        prefix = "".join([w[0].upper()for w in snake_name.split("_")])
```
gsantia, line 155:
```suggestion
prefix = "".join(w[0].upper() for w in snake_name.split("_"))
```

creator/ingest_runs/common/model.py
dankolbman: I'm also confused. Why are these old string formats being used instead of `ingest_processes` or something similar?

creator/ingest_runs/common/model.py
```diff
@@ -0,0 +1,182 @@
...
+    CANCELED = "canceled"
+    COMPLETED = "completed"
+
+
+class IngestProcess(models.Model):
+    """
+    Common model functionality for ingest processes
+    (e.g. ingest run, validation run)
+    """
+    __queue__ = "ingest"
```
dankolbman, line 46: Would this be better as an app setting? I think you also want to add the queue to `RQ_QUEUES`

creator/ingest_runs/common/model.py
```diff
@@ -0,0 +1,182 @@
...
+def hash_versions(versions):
+    """
+    Uniquely identify a set of file Versions by computing an MD5 hash digest
+    using the primary keys of the Versions
+    """
+    version_id_str = "".join(sorted(v.kf_id for v in versions))
+    return hashlib.md5(version_id_str.encode("utf-8")).hexdigest()
+
+
+class State(object):
```
dankolbman, line 33: Would this be better served by an Enum? We also tend to omit the `object` as it's already implied.

creator/ingest_runs/models/ingest_run.py
```diff
@@ -0,0 +1,56 @@
...
+        related_name="ingest_runs",
+        help_text="List of files to ingest in the ingest run",
+    )
+
+    @property
+    def study(self):
+        if self.versions.count() > 0:
+            return self.versions.first().root_file.study
+
+    def compute_name(self):
```
dankolbman, line 43: Why not make this a property? If there is a name, return it, if not, compute and save it then return it.

creator/ingest_runs/models/validation_run.py
```diff
@@ -0,0 +1,195 @@
...
+    )
+    data_review = models.ForeignKey(
+        DataReview,
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="validation_runs",
+    )
+
+    @property
+    def versions(self):
```
dankolbman, line 62: If there is a data review, you will return a queryset to fetch versions.
Maybe if there is not a data review, you could return `Version.objects.none()` so you have the same return type?

creator/ingest_runs/models/validation_run.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Get file versions from the associated DataReview
+        """
+        if self.data_review:
```
dankolbman, line 66: You may want to be careful with this as it may be pulling the full data review from the database. It could be expensive if there's ever a time when many validation runs get enumerated.

creator/ingest_runs/common/model.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def camel_to_snake(camel_str):
+    """
+    Convert CamelCase to snake_case
+    """
+    return re.sub(r'(?<!^)(?=[A-Z])', '_', camel_str).lower()
+
+
+def hash_versions(versions):
```
dankolbman, line 24: This requires a list of versions to be passed which may often require a bunch of unneeded fields (you only need the kf_id of each version).
You could consider accepting a list of kf_ids instead so you could use `values()` or `values_list()` to fetch only the kf_id field without having to pull the complete version row.

creator/ingest_runs/common/model.py
znatty22: Django's model inheritance rules say that you must specify a unique name for `related_name` so we can't just make it `ingest_processes` (and also I think that would be a little confusing since we have two different kinds of ingest processes). 

They do recommend using both the app label and the class name to guarantee uniqueness (see https://docs.djangoproject.com/en/3.1/topics/db/models/#be-careful-with-related-name-and-related-query-name) but I thought that this `User.ingest_runs_ingestruns` looks gross vs `User.ingestruns` which is more aligned with how we specify `related_name` on all our other models.

creator/ingest_runs/common/model.py
znatty22: Ehh, normally yes, but I use the State class a lot to refer to the states and I just hate having to do `State.RUNNING.value` vs `State.RUNNING` everywhere.

creator/ingest_runs/models/ingest_run.py
znatty22: I think I wanted the `name` field back when I was developing the ingest run model and needed it for debugging. I didn't make it a property bc I wanted it to be a field that got persisted in the db so I could filter by name in the queries. I think we could prob just get rid of it now?

creator/ingest_runs/models/validation_run.py
znatty22: Hmm I don't think so. I think that would only be true if you used `select_related` or if you actually iterate over the data review validation_runs queryset (see https://docs.djangoproject.com/en/3.1/ref/models/querysets/#when-querysets-are-evaluated)

creator/ingest_runs/common/model.py
dankolbman: I see. To get closer to what we use on other fields, maybe you want to use that handy `camel_to_snake()` here to turn it into `ingest_runs`?
Also, I think you have one too many `s` here.

creator/ingest_runs/models/ingest_run.py
dankolbman: Ah right. You could have the database field called `_name` and then use the `name` property to get and set that, but it's strange because you can use `name` everywhere except when you want to query on the db`.

Did we rule out using signals before to do this sort of thing? I think there's also a lifecycle hook you could compute and save the name in, too. Otherwise we have to remember to compute the name and save it every-time we instantiate an IngestRun.

creator/ingest_runs/models/validation_run.py
dankolbman: I'm not sure, but I'd want to test it out if it ever needs optimization. I'm suspicious of this pulling the whole row even if theoretically it only needs the foreign key column. I'd then wonder if maybe using `.exists()` would solve it or not.

creator/ingest_runs/models/validation_run.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            return Version.objects.none()
+
+    @property
+    def study(self):
+        """
+        Get Study from the associated data review
+        """
+        if self.data_review:
+            return self.data_review.study
```
dankolbman, line 77: Same thing here about return types. Here you could return `Study.objects.none()`.

dankolbman: Returning a queryset vs `None` is neat because you can treat things predictably.
Consider you want to do:
```
validation_run.study.collaborators.all()
```
This is fine if the study exists, but if you return `None`, then you break the API that the ORM provides and have to handle things differently in your code.
If you return an empty queryset, I believe you can still try to access collaborators, but operating on an empty queryset should just result in another empty queryset. If this isn't the case, then you would at least get a more formal `Study.DoesNotExist` or similar error.

creator/ingest_runs/models/validation_run.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+    did_not_run = models.IntegerField(
+        null=True,
+        blank=True,
+        default=0,
+        help_text="The number of validation tests that did not run"
+    )
+
+    @property
+    def study(self):
```
dankolbman, line 164: Same thing here about returning querysets.

creator/ingest_runs/common/model.py
znatty22: Lol that's what I thought too with that extra `s`, but `%(class)s` gets replaced with `ingestrun` so you need the extra `s` for it to be `ingestruns`

creator/ingest_runs/models/ingest_run.py
znatty22: Yea we already do that here: https://github.com/kids-first/kf-api-study-creator/blob/8f8ee36693b9c2c2ff4110715f0c331b38d62605/creator/ingest_runs/signals.py#L22

creator/ingest_runs/models/validation_run.py
znatty22: Ok let me do some testing to see

creator/ingest_runs/models/validation_run.py
znatty22: Ah yea, I meant to but forgot to return the empty queryset in a couple places

creator/ingest_runs/common/model.py
znatty22: I'm not sure how easy it is or if its possible to set `related_name` with a function..I'd have to look into it

creator/ingest_runs/models/validation_run.py
znatty22: I think we're ok here. So here are all of the queries that happen when you try to access `data_review` on a `ValidationRun`:

```
In [2]: from creator.ingest_runs.models.validation_run import ValidationRun

In [3]: vr_has_dr = ValidationRun.objects.get(data_review__isnull=False)
SELECT "ingest_runs_validationrun"."id",
       "ingest_runs_validationrun"."created_at",
       "ingest_runs_validationrun"."creator_id",
       "ingest_runs_validationrun"."input_hash",
       "ingest_runs_validationrun"."state",
       "ingest_runs_validationrun"."job_log_id",
       "ingest_runs_validationrun"."success",
       "ingest_runs_validationrun"."progress",
       "ingest_runs_validationrun"."data_review_id"
  FROM "ingest_runs_validationrun"
 WHERE "ingest_runs_validationrun"."data_review_id" IS NOT NULL

Execution time: 0.002568s [Database: default]

In [4]: vr_has_dr.data_review
SELECT "data_reviews_datareview"."kf_id",
       "data_reviews_datareview"."uuid",
       "data_reviews_datareview"."created_at",
       "data_reviews_datareview"."creator_id",
       "data_reviews_datareview"."name",
       "data_reviews_datareview"."description",
       "data_reviews_datareview"."state",
       "data_reviews_datareview"."study_id"
  FROM "data_reviews_datareview"
 WHERE "data_reviews_datareview"."kf_id" = 'DR_D2GE4DV6'

Execution time: 0.006991s [Database: default]
Out[4]: <DataReview: DataReview object (DR_D2GE4DV6)>
```

It does pull the whole row, but I don't think this matters much since the fields are just a bunch of strings essentially. The important part is it doesn't pull in any of data_review's related objects

dankolbman (approved):
# 🧱

PR #610, opened 2021-03-30
🔥 Remove lingering ego auth fields

dankolbman:
Removes the `ego_roles` and `ego_groups` fields from the `User` model.

Needed to also update the subscription mutations to rely on the Django permission models instead of ego.

Checks around empty profiles and services will be cleaned up in a later PR addressing service tokens.


XuTheBunny (approved):
🍟

znatty22 (approved):
✅

gsantia (approved):

PR #612, opened 2021-03-31
✨ Save services as users

### LINKED ISSUE #597: Better support for services
Services should be able to authenticate through Auth0 m2m tokens using the `client_credentials` flow.
We should register new 'service' users whenever we see a new a sub come in and allow these special users to be configured through an admin page.

We should allow modification of basic attributes such as:
- username
- avatar
- permission groups

dankolbman:
Creates a new user for services when a service token is used as authentication.
Requires that the application that is associated with the `AUTH0_SECRET` and `AUTH0_CLIENT` pair be authorized for the Auth0 Management API and have the `read:clients` scope.
We may want to make this more adaptable to more auth backends some time in the future. It's not ideal that we rely on the Auth0 API to fetch more info about the service.

Fixes #597 

Waiting on #610


creator/middleware.py
```diff
@@ -124,10 +114,15 @@ def get_jwt_user(request):
             # The user is already in the database, update their last login
             update_last_login(None, user)
         except User.DoesNotExist:
-            profile = Auth0AuthenticationMiddleware._get_profile(encoded)
+            profile = None
+            if token.get("gty") == "client-credentials":
```
znatty22, line 118: Just a question, so can service users just send any query to get registered?

znatty22 (approved):
💰

PR #613, opened 2021-04-06
✨ Validation GraphQL API

### LINKED ISSUE #551: Validation Job and GraphQL API
## Overview

Related to #550 

We need some GraphQL queries and mutations to allow Data Tracker users to start and cancel validation jobs and fetch validation results during a Data Review. Validation runs/jobs should probably never be deleted since they serve as a method of data provenance for the study data.

## Tasks - API

Depending on how #550 is implemented, the GraphQL API for a validation run may simply entail modifying the current IngestRun API so that the correct validation parameters can be supplied.

- [ ] Modify or add new mutation to start a job which runs validation on the data review documents
- [ ] Modify or add new mutation to cancel a running validation job
- [ ] Add query to fetch validation runs, filter params TBD.
- [ ] Maybe need a "special" query to fetch validation results (they tend to be big)

## Tasks - Job

Validation jobs should only run if the input files adhere to the expedited file types. We may need to do some conversion between the human friendly column names in the expedited file types and the concept names expected by the validation library.

- [ ] Modify the start_ingest_run task or create a new task to run validation on a set of file versions
- [ ] More stuff, but TBD for now

znatty22:
GraphQL API to start/cancel validation runs and view a data review's validation results. Includes a placeholder for file validation which stores dummy validation result summary on the data review (DataReview.validation_resultset) and returns a sample validation report. 

Closes #551


creator/ingest_runs/queries/validation_run.py
```diff
@@ -0,0 +1,37 @@
+from graphene import relay
+from graphene_django.filter import DjangoFilterConnectionField
+from graphql import GraphQLError
+from django_filters import FilterSet, OrderingFilter
+
+from creator.ingest_runs.nodes import ValidationRunNode
+from creator.ingest_runs.models import ValidationRun
+
+
+class ValidationRunFilter(FilterSet):
```
dankolbman, line 10: Do we ever need to filter by study or data review?

dankolbman (approved):

creator/ingest_runs/queries/validation_run.py
znatty22: Maybeee? If we had a view that listed _all_ validation runs then these two filters would be useful. But right now all we have is a detail view for a data review where would likely display a list of validation runs for that review by accessing `DataReview.validation_runs`.

tests/ingest_runs/test_common.py
```diff
@@ -85,9 +93,54 @@ def test_ingest_process_states(
...
+    for i in range(2):
+        study_id, file_id, version_id = prep_file(authed=True)
+        versions.append(Version.objects.get(pk=version_id))
+    process = factory(versions=versions[0:1], state=state)
+
+    # Ingest processes with same set of versions are duplicates and should be
+    # canceled
+    canceled_any = cancel_duplicate_ingest_processes(
+        versions[0:1], process.__class__, cancel_task
+    )
```
gsantia:
```suggestion
version_subset = versions[:1]
process = factory(version_subset, state=state)

# Ingest processes with same set of versions are duplicates and should be
# canceled
canceled_any = cancel_duplicate_ingest_processes(
    version_subset, process.__class__, cancel_task
)
```

tests/ingest_runs/test_ingest_runs_mutations.py
```diff
@@ -150,7 +139,12 @@ def test_start_duplicate_run(
     # the previous IngestRun to get canceled.
     resp2 = send_query(client, START_INGEST_RUN, {"versions": version_ids})
     # Check that the correct enqueue call occurred.
-    mock_cancel_duplicate.assert_called_once()
+
+    # Check that cancel duplicates was called and run_ingest was not called
+    mock_ingest_queue.enqueue.call_count == 0
+    mock_cancel_duplicates.call_count == 1
+    args, _ = mock_cancel_duplicates.call_args_list[0]
+    assert set(args[1:]) == set([IngestRun, cancel_ingest])
```
gsantia, line 149:
```suggestion
assert set(args[1:]) == {IngestRun, cancel_ingest}
```

tests/ingest_runs/test_validation_result_downloads.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    file_field.save("data.csv", File(open("tests/data/data.csv")))
+    assert file_field.storage.exists(file_field.name)
+
+    # Download file
+    client = clients.get("Administrators")
+    download_url = data_review.validation_resultset.report_path
+    resp = client.get(download_url)
+
+    resp.status_code == 200
```
gsantia, line 101:
```suggestion
    resp.status_code == 200
```

gsantia (approved):

PR #617, opened 2021-04-13
✨ Data review file validation task

### LINKED ISSUE #623: Implement file validation task
Replace the placeholder logic in the `creator/ingest_runs/tasks/validation_run.py` with real business logic that executes file validation and persists results.

### LINKED ISSUE #636: Add intermediate states to ingest processes
When a user clicks the button to cancel a run, there is a noticeable delay between the request to cancel the run and the completion of the cancellation since canceling requires queuing a background task which doesn't start right away. There are similar issues with starting a validation run run. 

### TODO:

We should add a `cancelling` state to the `IngestProcess` model (which is inherited by validation run model). The `IngestProcess` state transition methods will need to change a bit:

- [ ] Replace the `IngestProcess.cancel` method with `IngestProcess.start_cancel` and `IngestProcess.complete_cancel`
- [ ] Update the cancel mutations for both ingest run and validation run to call the start cancel method before the cancel task is queued
- [ ] Update the cancel tasks for both ingest run and validation run to call the complete cancel method instead of the old cancel method

We should add an `initializing` state to `IngestProcess` model (which is inherited by validation run model). The `IngestProcess` state transition methods will need to change a bit:

- [ ] Replace the `IngestProcess.start` method with `IngestProcess.initialize` and `IngestProcess.start`
- [ ] Update the start mutations for both ingest run and validation run to call the initialize method before the start task is queued
- [ ] Update the start tasks for both ingest run and validation run to call the initialize method instead of the old start method

### LINKED ISSUE #637: Capture reason for failed ingest processes
When a validation run fails an exception is raised with a useful error message, however this message is not stored or sent back to the frontend since the exception happens in an async worker. This results in the user not knowing why a run failed. 

We should add an `error_msg` field to the `IngestProcess` model (this gets inherited by `IngestRun` and `ValidationRun`) that will get populated in the ingest process task.

### Todo:

- [x] Add `IngestProcess.error_msg`
- [x] Update the `run_ingest` task and `run_validation` task to populate this field in the exception handler before the fail method is called on the run
- [x] Update tests to reflect this new field

### LINKED ISSUE #640: Clear validation results when appropriate
**Clear results on file changes**
A user runs a validation on the review files and the report is displayed. Then the user removes all of the files from the review. Right now the validation results from the previous run is still displayed. We should clear the review's validation result whenever the data review files change.

**Clear results on start of new validation**
A user runs a validation on the review files and the report is displayed. A user re-runs validation but the previous report is still displayed. We should always clear out the results when a new validation run is started. Even though the files being validated might not have changed, the validation rules could have changed.

znatty22:
Closes #623 
Closes #636 
Closes #640 
Closes #637

- Replaces the placeholder validation task with real business logic that performs validation on a data review's file versions and persists the results. Also includes validation run cancellation handlers for when a version or data review is deleted.
- Adds intermediate states `initializing` and `canceling` to ingest processes
- Clears validation results when data review files change and when a validation run is started
- Capture summary of ingest process exceptions on each ingest process instance

**Updated ValidationRun FSM**

image


creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,159 @@
...
+    logger.info(f"Performing extraction to clean and map file {version}")
+
+    # Load ingest extract operations for this file
+    if not version.root_file:
+        raise Exception(
+            f"Validation failed. Version {version} missing root_file"
+        )
+    ec = extract_config_path(version)
+    if not ec:
+        raise Exception(
```
znatty22, line 49: I'm not sure what the desired behavior is here. Is it possible to have a file in a data review that does not have a template? If yes, then I would change this to log a warning instead of raise an Exception

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,163 @@
...
+            "⚠️  Could not find an extract config for file version {version} "
+            "with file type: {version.root_file.file_type}. This file will "
+            "not be included in validation."
+        )
+        return None
+    # Extract file content into a DataFrame
+    try:
+        df = pandas.DataFrame(extract_data(version))
+    except Exception as e:
+        logger.error(
```
gsantia, line 59:
```suggestion
        logger.exception(
```

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,163 @@
...
+        .only("root_file__file_type")
+        .filter(kf_id__in=set(results["files_validated"])).all()
+    )
+    for version in versions:
+        fname = FILE_TYPES[version.root_file.file_type]["name"]
+        friendly_names.append(f"{fname}, {version.kf_id}")
+    results["files_validated"] = friendly_names
 
+    logger.info(
+        f"Finished validation, building report for: pformat({friendly_names})"
```
gsantia, line 118:
```suggestion
        f"Finished validation, building report for {pformat(friendly_names)}"
```

tests/ingest_runs/test_validation.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    Test file validation and report building
+    """
+    versions = data_review_with_files.versions.all()
+    vr = ValidationRunFactory(data_review=data_review_with_files)
+    results = validation_run.validate_file_versions(vr)
+
+    # Check that all versions for the review were validated
+    assert results
+    assert set(results["files_validated"]) == set({v.kf_id for v in versions})
```
gsantia, line 129:
```suggestion
    assert set(results["files_validated"]) == {v.kf_id for v in versions}
```

tests/ingest_runs/test_signals.py
```diff
@@ -10,6 +15,75 @@
...
+    for i in range(2):
+        if i == 0:
+            state = State.RUNNING
+        else:
+            state = State.FAILED
+        vrs.append(
+            ValidationRunFactory(
+                data_review=data_review, state=state, creator=user
+            )
+        )
```
gsantia, line 39:
```suggestion
    vrs = [ValidationRunFactory(data_review=data_review, state=x, creator=user) for x in (State.RUNNING, State.FAILED)]
```

tests/ingest_runs/test_signals.py
```diff
@@ -10,6 +15,75 @@
...
+    versions = []
+    for i in range(2):
+        _, _, version_id = prep_file(authed=True)
+        versions.append(Version.objects.get(pk=version_id))
+    vrs = []
+    drs = []
+    for i in range(3):
+        if i == 2:
+            dr = DataReviewFactory(versions=versions[1:])
+            vr = ValidationRunFactory(state=State.RUNNING, creator=user)
```
gsantia, line 69:
```suggestion
            vr = ValidationRunFactory(
                state=State.RUNNING, data_review=dr, creator=user
            )
````

File: creator/ingest_runs/tasks/validation_run.py (hunk 2 above)
Comment (znatty22) [reply creator/ingest_runs/tasks/validation_run.py:60]: @gsantia If I use logger.exception, the my message and the exception traceback will be logged. The task decorator already automatically logs the exception traceback so that's why I used logger.error here and re-raised the exception.

File: tests/ingest_runs/test_signals.py [hunk 7]
```diff
@@ -10,6 +10,70 @@
...
+        if i == 2:
+            dr = DataReviewFactory(versions=versions[1:])
+            vr = ValidationRunFactory(
+                state=State.RUNNING, data_review=dr, creator=user
+            )
+        else:
+            dr = DataReviewFactory(versions=versions[0:1])
+            vr = ValidationRunFactory(
+                state=State.RUNNING, data_review=dr, creator=user
+            )
```
gsantia, line 69:
```suggestion
        vers = versions[1:] if i ==2 else versions[:1]
        dr = DataReviewFactory(versions=vers)
        vr = ValidationRunFactory(
             state=State.RUNNING, data_review=dr, creator=user
        )
```

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,181 @@
...
+    """
+    logger.info(f"Performing extraction to clean and map file {version}")
 
-    Upload validation results and report files to storage backend
+    # Load ingest extract operations for this file
+    if not version.root_file:
+        raise Exception(
+            f"Validation failed. Version {version} missing root_file"
+        )
+    ec = extract_config_path(version)
```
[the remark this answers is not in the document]
znatty22: `extract_config` is just shorter, but misleading because its returning a path to the extract config not the extract config object itself

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,181 @@
...
+def validate_file_versions(validation_run):
+    """
+    Load ValidationRun.versions into DataFrames, extract data from each
+    version, clean and map the data before passing it to the validator
     """
-    logger.info(f"Begin validation for {validation_run.versions.all()}")
-    results = {}
+    versions = validation_run.versions.all()
+    logger.info(
+        f"Begin validating file versions: {pformat(list(versions))}"
```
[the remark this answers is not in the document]
znatty22: I guess I'd have to test it out to verify this. But I think when your list longer than some length, vanilla printing it will wrap the items to the next line, where as  if you use pformat, it'll print each item on a new line. I think the latter is more readable in logs.

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,79 +1,181 @@
...
+        )
+        raise
+
+    logger.info(
+        f"Applied extract config {os.path.split(ec)[-1]} for file "
+        f"{version.root_file.file_type}, {version.kf_id}"
+    )
+
+    ext = Extractor()
+    ext.logger = logger
```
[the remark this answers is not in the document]
znatty22: These logger hacks were workarounds bc the logging in `Extractor` and `MarkdownReportBuilder` don't work in study creator (log statements are not showing up). The reason is bc those loggers are created in the class constructor but should be created at the module level so we can add them in django logging settings. I'm going to have to make the fix in ingest lib before I can change it here.

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -85,33 +175,65 @@ def persist_results(results, report_md, validation_run):
...
 
     return resultset
 
 
 @task("run_validation")
 def run_validation(validation_run_uuid=None):
     """
-    TODO
-    Run validation process for a given set of files.
+    Run validation process for the file version in a ValidationRun
```
gsantia, line 240:
```suggestion
      Run validation process for the file versions in a ValidationRun
```

tests/ingest_runs/test_validation.py
```diff
@@ -0,0 +1,242 @@
...
+    mocker.patch(
+        "creator.ingest_runs.tasks.validation_run.validate_file_versions",
+        side_effect=Exception
+    )
+    vr = ValidationRunFactory(data_review=data_review)
+    with pytest.raises(Exception):
+        validation_run.run_validation(str(vr.pk))
+        vr.refresh_from_db()
+        assert not vr.success
+        assert vr.state == State.FAILED
```
gsantia: These assertions will pass no matter what since they're inside the raises block
```suggestion
with pytest.raises(Exception):
        validation_run.run_validation(str(vr.pk))        
        vr.refresh_from_db()
assert not vr.success
assert vr.state == State.FAILED
```

tests/ingest_runs/test_validation.py
```diff
@@ -0,0 +1,242 @@
...
+        assert FILE_TYPES[ft]["name"] in report
+
+    # Test validation when all versions are missing extract cfgs
+    mock_clean_map = mocker.patch(
+        "creator.ingest_runs.tasks.validation_run.clean_and_map",
+        return_value=None
+    )
+    with pytest.raises(Exception) as e:
+        validation_run.validate_file_versions(vr)
+        assert "No templates were found" in str(e)
```
gsantia:
```suggestion
    with pytest.raises(Exception) as e:
        validation_run.validate_file_versions(vr)
    assert "No templates were found" in str(e)
```

gsantia:
```suggestion
    with pytest.raises(Exception) as e:
        error = e
        validation_run.validate_file_versions(vr)
    assert "No templates were found" in str(error)
```

tests/ingest_runs/test_validation.py
```diff
@@ -0,0 +1,242 @@
...
+        def __init__(self, root_file=None, extract_config_path=None):
+            self.root_file = root_file
+            self.extract_config_path = extract_config_path
+
+    # Version missing a root file
+    clean_and_map = validation_run.clean_and_map
+    mock_version = MockVersion()
+    with pytest.raises(Exception) as e:
+        clean_and_map(mock_version)
+        assert "missing root_file" in str(e)
```
gsantia:
```suggestion
    with pytest.raises(Exception) as e:
        clean_and_map(mock_version)
        error = e
    assert "missing root_file" in str(error)
```

tests/data_reviews/test_data_reviews_mutations.py
```diff
@@ -263,8 +265,13 @@ def test_update_data_review(
         assert resp.json()["errors"][0]["message"] == "Not allowed"
         return
 
-    # Check expected end_state
     data_review.refresh_from_db()
+
+    # Check that validation results were cleared
+    with pytest.raises(ValidationResultset.DoesNotExist):
+        assert data_review.validation_resultset
```
gsantia, line 272:
```suggestion
    with pytest.raises(ValidationResultset.DoesNotExist):
        temp = data_review.validation_resultset
```

tests/ingest_runs/test_validation_runs_mutations.py
```diff
@@ -98,6 +102,10 @@ def test_start_validation_run(
         assert vr["state"] == State.INITIALIZING
         assert data_review.kf_id == vr["dataReview"]["kfId"]
 
+        # Check that validation results were cleared
+        with pytest.raises(ValidationResultset.DoesNotExist):
+            assert data_review.validation_resultset
```
gsantia, line 107: same as above

tests/data_reviews/test_data_reviews_mutations.py
gsantia, line 272: not entirely sure what to suggest, but I think this does it?

```suggestion
    with pytest.raises(ValidationResultset.DoesNotExist):
        _ = data_review.validation_resultset
```

tests/ingest_runs/test_validation_runs_mutations.py
gsantia, line 107:
```suggestion
        with pytest.raises(ValidationResultset.DoesNotExist):
            data_review.validation_resultset
```

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -94,6 +91,17 @@ def validate_file_versions(validation_run):
...
+
+    # TEMPORARY FOR TESTING
+    import time
+
+    t_end = time.time() + 20
+    progress = ""
+    while time.time() < t_end:
+        time.sleep(1)
+        progress += "."
+        logger.info(progress)
```
gsantia: still want this?

znatty22: Once we're finished testing with frontend I'll remove this

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,82 +1,171 @@
...
+            f"Validation failed. Version {version} missing root_file"
+        )
+    ec = version.extract_config_path
+    if not ec:
+        logger.warning(
+            f"⚠️  Could not find an extract config for file version {version} "
+            f"with file type: {version.root_file.file_type}. This file will "
+            "not be included in validation."
+        )
+        return None
```
dankolbman, line 51: Why not raise here instead?

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,82 +1,171 @@
...
+    version, clean and map the data before passing it to the validator
     """
-    logger.info(f"Begin validation for {validation_run.versions.all()}")
-    results = {}
+    versions = validation_run.versions.all()
+    logger.info(f"Begin validating file versions: {pformat(list(versions))}")
+    # Clean and map source data
+    df_dict = {}
+    for version in versions:
+        clean_df = clean_and_map(version)
```
dankolbman, line 83: No error handling?
Maybe it would be better to perform the skip logic here. If there's a problem trying to clean or map, throw a warning and skip it.

dankolbman (changes requested):
Can you make sure to run everything through black? Looks like there's some dangling style issues.

creator/files/models.py
```diff
@@ -326,6 +329,23 @@ def path(self):
...
+    @property
+    def extract_config_path(self):
+        """
+        Get the extract config file path for the Version
+        """
+        config_path = None
+        try:
+            file_type = self.root_file.file_type
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
znatty22: I think this would happen if there's no root_file

creator/ingest_runs/tasks/validation_run.py
znatty22: It depends on what kind of behavior we want. Should we fail validation if one or more of the files in the review doesn't have an extract config? Or should we allow validation to continue for the files that do have extract configs?

creator/ingest_runs/tasks/validation_run.py
znatty22: I would rather raise the Exceptions and log the warnings inside the `clean_and_map` method bc there are several different error cases and messages that have specific messages

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -1,82 +1,171 @@
...
+    if not df_dict:
+        raise Exception(
+            "Validation failed! Validation only runs on files that conform "
+            "to expedited file types. None of the input files conformed to "
+            "any of the available expedited file types"
+        )
 
-    return results
+    # Validate
+    return DataValidator().validate(df_dict, include_implicit=True)
```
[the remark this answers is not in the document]
znatty22: I added type annotations to all of the methods in this module

creator/files/models.py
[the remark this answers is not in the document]
dankolbman: Can we narrow it down to `File.DoesNotExist` then?

creator/ingest_runs/tasks/validation_run.py
dankolbman: To me, the purpose of this function is to clean and map a data file. If an extract config cannot be found, then it cannot complete its purpose and fails. Similarly, if something goes wrong with extracting the data, it also fails. However, one case is encoded in an exception and the other is encoded by the `None` value. 
If you instead were to raise exceptions specific to the problem, the calling context (`validate_file_versions()`) would be able to tell why exactly it failed and act accordingly (continue validation if no extract config, fail out if extract fails). Furthermore, a developer reading the `validate_file_versions()` function would immediately understand exactly what case `except ExtractConfigDoesNotExist` is handling without having to read through `clean_and_map()` to understand why `None` is being returned.

creator/ingest_runs/tasks/validation_run.py
dankolbman: I don't think you need to handle all exceptions here, just ones specific to determining whether to fail out or to skip a file.
This is the place where you're ultimately deciding which path to take, which is currently done with the `isinstance` check, but consider how something like this may be more readable and generalizable to other failure cases:
```python
for version in versions:
    try:
        df_dict[version.kf_id] = clean_df
    except ExtractConfigDoesNotExist:
        logger.warning("Extract configuration does not exist. Will not try to validate document")
        continue
    except Exception as err: # It would be better to have some sort of exception specific to extract_data here
        logger.error("Something went wrong trying to extract file contents")
        raise
```

creator/files/models.py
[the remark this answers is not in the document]
znatty22: Fixed. Although I had to use `AttributeError`. It gets raised bc `root_file` is a `NoneType` when its not set

creator/ingest_runs/tasks/validation_run.py
znatty22: Oh I think I misunderstood your original comment. I agree with your feedback. Updated

dankolbman (approved):
# :star2:

PR #618, opened 2021-04-14
📝 Add RFC for Organization

dankolbman:
RFC write up for the proposed Organization entity.


docs/rfcs/0001_organizations.md
```diff
@@ -0,0 +1,98 @@
+# Organizations
+
+**Status:** Approved, waiting for implementation
```
dankolbman, line 3: This will be 'approved' once it merges, until then, it's still in review.

docs/rfcs/0001_organizations.md
```diff
@@ -0,0 +1,98 @@
...
+
+An organization encompasses a group of people working together to integrate many studies.
+
+## Why
+
+The concept of an an Organization arises from the need to manage groupings of studies and people with similar tooling. With an Organization, an institution or data coordinating center can establish a workspace for itself and administer its own studies and personnel. An Organization is intended to separate administrative needs from the data collection process.
+
+## How
+
+An Organization encompasses two important entities: studies and users. An Organization owns many studies where each study is related strictly to one Organization. An Organization also contains many users although each user may be a member of many Organizations.
```
znatty22, line 13: > An Organization encompasses two important entities: studies and users. 

Is storage backends not included here because it will be added in another RFC?

docs/rfcs/0001_organizations.md
```diff
@@ -0,0 +1,98 @@
...
+##### Schema
+```graphql
+type CreateOrganizationInput {
+  name: String!
+}
+createOrganization(input: CreateOrganizationInput): Organization!
+```
+##### Permissions
+
+Any authenticated user may create a new organization.
```
znatty22, line 49: Hmm maybe only superusers (I think that's what our current Admin group represents) should be able to create a new organization?

docs/rfcs/0001_organizations.md
```diff
@@ -0,0 +1,98 @@
...
+- allStudies(name: String): StudyNode!
++ allStudies(name: String, organization: ID!): StudyNode!
+```
+```diff
+- Study.objects.filter(deleted=False).all()
++ organization = Organization.objects.get(organization_id)
++ Study.objects.filter(deleted=False, organization=organization).all()
+```
+### Permissions
+Broad permission groups such as `Administrators` and `Investigators` are applied application-wide. These scopes will be too large with the introduction of Organizations. Permissions will likely need to be controlled per-Organization such as a user being Administrator in Organization A but only Investigator in Organization B. This will require some refactor of the permissions mechanism.
```
znatty22, line 97: I wonder if it would still be useful to have a superuser group that would have admin privs for all organizations.

znatty22 (commented):
There isn't any mention of app routes changing or use of subdomains for organization so I'm assuming we're keeping those the same? So for example, the route for a file will be `/studies/SD_ME0WME0W/documents/SF_00000000` vs something like `/kids-first/studies/SD_ME0WME0W/documents/SF_00000000` or `kids-first.kf-ui-data-tracker.org/studies/SD_ME0WME0W/documents/SF_00000000`

docs/rfcs/0001_organizations.md
znatty22: Also wondering if we should support templates per organization or per study?

dankolbman: Storage backends, templates, and whatever else are future propositions whereas studies and users are things we have to deal with at implementation time for Organizations. Those details can be spec'd in their own docs.

docs/rfcs/0001_organizations.md
dankolbman: Perhaps to start, but the idea is that anyone may join and create an organization.

docs/rfcs/0001_organizations.md
dankolbman: Yeah the permissions are going to need their own re-design.

znatty22 (approved):
👥 ✅

PR #620, opened 2021-04-16
🔊 Warn instead of raising exception

dankolbman:
Logs a warning message instead of raising an exception when one or more errors are encountered when analyzing files.
This avoids trackbacks in logs and errors appearing on Sentry when documents fail analysis.

### DIFF
--- a/creator/tasks.py
+++ b/creator/tasks.py
@@ -254,6 +254,7 @@ def analyzer_task():
             analysis = analyze_version(version)
         except Exception:
             errors += 1
+            logger.warning(f"Failed to analyze version {version.kf_id}")
             continue
         analysis.user = version.creator
         analysis.save()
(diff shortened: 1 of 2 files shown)


creator/tasks.py
```diff
@@ -261,5 +261,3 @@ def analyzer_task():
         f"Successfully analyzed {len(versions) - errors} versions. "
         f"Failed to analyze {errors} versions."
     )
-    if errors > 0:
```
znatty22, line 264: Is this supposed to be replaced with a logger warning?

gsantia:
```suggestion
    if errors > 0:
        logger.warning(f"Failed to analyze {errors} versions.")
```

creator/tasks.py
```diff
@@ -261,5 +261,3 @@ def analyzer_task():
...
-        raise Exception(f"Failed to analyze {errors} versions.")
```
gsantia, line 265:
```suggestion
    if errors > 0:
        logger.warning(f"Failed to analyze {errors} versions.")
```

gsantia (changes requested):
spelling error in the commit message, otherwise ✅

gsantia (approved):

znatty22 (approved):

PR #628, opened 2021-04-23
✨ Add debug middleware

dankolbman:
Adds a `_debug` field in the query schema so that we may see raw sql executed for each query.
This middleware is only added for the development settings and requires `DEBUG = True` to be added to the schema.

**Example:**
```
query Study {
  _debug {
    sql {
      rawSql
      duration
    }
  }
  status {
    name
    version
    commit
    queues
  }
}
```

### DIFF
--- a/creator/schema.py
+++ b/creator/schema.py
@@ -6,6 +6,7 @@ or Mutation from the application's schema module.
 No resolvers or type definitions should be included here.
 """
 import graphene
+from django.conf import settings
 
 import creator.analyses.schema
 import creator.buckets.schema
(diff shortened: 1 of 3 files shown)


znatty22 (approved):
Maybe you've already tried these but just in case... Have you tried django shell_plus? It has a bunch of useful features and one lets you view SQL queries in the django shell in real time as you play around with the ORM: https://django-extensions.readthedocs.io/en/latest/shell_plus.html#sql-queries

Or even just printing the SQL query from a QuerySet (`print(Version.objects.all().query)`) has been pretty helpful for me when optimizing queries.

gsantia (approved):

dankolbman:
@znatty22 yeah I've seen the extensions before but usually I'm fine with just accessing the queries directly from the connection or in this case, doing it from a test. The test approach will also ensure that we don't accidentally introduce something that hurts a heavily used query, too.

Issue #619, opened 2021-04-15
Improve study query performance


dankolbman:
The `study` and `allStudies` queries are slow as they are performing many sequential SQL queries.
Using data loaders or prefetching relationships help reduce some of the linear scaling in the queries, but there are still issues trying to reduce n+1 on the `Membership` fields when resolving collaborators.

## Current Performance

<img width="798" alt="Screen Shot 2021-04-15 at 9 52 34 AM" src="https://user-images.githubusercontent.com/2495894/114880780-5cc36600-9dd0-11eb-9f43-d798cb757d89.png">
<img width="789" alt="Screen Shot 2021-04-15 at 9 53 25 AM" src="https://user-images.githubusercontent.com/2495894/114880856-71076300-9dd0-11eb-92f8-6fb80fa7ee90.png">

PR #629, opened 2021-04-23
⚡ Optimize study queries

dankolbman:
Adds data loaders to reduce the number of sql queries being sent to the database:
- `creators` on releases
- add `latestRelease` to replace `allStudies { releases() }` subquery

Reduces what was originally 41 queries down to 18. Still a lot of work that could be done.

Addresses #619


znatty22:
Closing due to inactivity

PR #639, opened 2021-05-06
✨ Add additional fields to study model

### LINKED ISSUE #635: Add additional fields to study model
We should add a new field on the `Study` model that will capture any additional, non-standard fields for each study.
For example, we may wish to link out to a dbGaP page for the study, a jira ticket, or list additional buckets that are used by the study.
This field should be called `additional_fields` and be a `JSONField`. This field will store data generated on the front end using [Editor.js](https://editorjs.io/) which produces 'blocks' inside of a JSON object.

Addresses #324

XuTheBunny:
Screen Shot 2021-05-07 at 1 15 53 PM

Closes #635


dankolbman (approved):
# 🥽

PR #643, opened 2021-05-11
✨Add error_msg field in IngestProcess

### LINKED ISSUE #637: Capture reason for failed ingest processes

gsantia:
Closes #637.

Add an `error_msg` field to the `IngestProcess` model, update the `run_ingest` and `run_validation` tasks to populate this field when exceptions are raised, and update the tests to reflect these changes.


creator/ingest_runs/migrations/0004_auto_20210511_1505.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.20 on 2021-05-11 15:05
```
znatty22, line 1: Can you delete this migration file and regenerate it but this time supply the name (maybe `add_ingest_error_msg` for the migration?

creator/ingest_runs/tasks/ingest_run.py
```diff
@@ -31,28 +31,36 @@ def run_ingest(ingest_run_uuid=None):
...
+            ingest_run.fail()
+            ingest_run.save()
+            raise Exception(feature_msg)
         logger.info(
             "All files in the ingest run are genomic workflow manifests."
         )
         try:
             ingest_genomic_workflow_output_manifests(ingest_run)
         except Exception:
+            ingest_run.error_msg = (
```
znatty22, line 51: This is going to overwrite the message you set above.

creator/ingest_runs/tasks/ingest_run.py
```diff
@@ -31,28 +31,36 @@ def run_ingest(ingest_run_uuid=None):
...
         if not settings.FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS:
-            ingest_run.fail()
-            ingest_run.save()
-            raise Exception(
-                "Ingesting genomic workflow output manifests is not enabled."
+            feature_msg = (
+                "Ingesting genomic workflow output manifests is not enabled. "
                 "Make sure that FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS is set."
             )
+            ingest_run.error_msg = feature_msg
```
znatty22, line 41: I'm wondering if there's a better way to do this. So instead of setting the error message in all the places in the code where an exception is about to be raised, could the error message just be set in the top level ingest process exception handler using the exception object? No idea what this will give us so I'll leave it up to you to figure out, but something like this? 

```
except Exception as e:
    logger.error(str(e))
    ingest_run.error_msg(str(e))
    ingest_run.fail()
    ingest_run.save()
```

The same approach could be used in the validation task too.

znatty22 (changes requested):

creator/ingest_runs/tasks/ingest_run.py
```diff
@@ -31,28 +31,33 @@ def run_ingest(ingest_run_uuid=None):
     are_gwo = all(check_gwo(version) for version in versions)
     if are_gwo:
```
znatty22, line 32: Could this block of code be refactored slightly so you have only one place where you're calling ingest_run.fail?

```python
        try:
            if are_gwo:
                if not settings.FEAT_INGEST_GENOMIC_WORKFLOW_OUTPUTS:
                    raise Exception(.. the feature not enable msg ..)
                else:
                    ingest_genomic_workflow_output_manifests(ingest_run)
            else:
                raise Exception(...the unknown file msg..)
        except Exception as e:
            ingest_run.fail(str(e))
            ingest_run.save()
            raise
```

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -249,15 +248,17 @@ def run_validation(validation_run_uuid=None):
...
         report_markdown = build_report(results)
         vrs = persist_results(results, report_markdown, vr)
-    except Exception:
+    except Exception as e:
         vr.success = False
-        vr.fail()
+        vr.fail(error_msg=str(e))
         vr.save()
-        raise
+        er_msg = "Error occurred during validation run process."
```
znatty22, line 260: Isn't this going to overwrite whatever message is in `e`? I think you just need to replace 

```python
er_msg = "Error occurred during validation run process."
raise type(e)(er_msg).with_traceback(sys.exc_info()[2])
```
with `raise`

znatty22 (changes requested):

znatty22 (approved):
👌

PR #647, opened 2021-05-17
📝 Data Submission Templates RFC

znatty22:
RFC write up for the proposed data submission templates


docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,347 @@
+# 🗃 Data Submission Templates
+
+**Status**: Approved, waiting for implementation
+
+This document expands on the [MVP solution](0002_templates_mvp.md#motivation)
+by adding new concepts such as template versioning.
+
+## 📓 Model
+
+![Relational Diagram](../_static/images/template-models.png)
```
dankolbman, line 10: I would suggest implementing the whole model at once as adding the versions later may make migrations tricky.

dankolbman: I think you may want a one-to-one from study to PackageVersion rather than a TemplatePackage. Otherwise, how will you know which version the study is using (unless they always use the latest)?

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+    study: ID
+}
+```
+
+### UpdateTemplatePackageMutation
+
+Updates a TemplatePackage and optionally assigns it to a study.
+
+```graphql
+type UpdateTemplatePackageInput {
```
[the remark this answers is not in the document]
znatty22: Yea, I omitted the mutation definition for brevity

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+
+In order for a template to be used in studies, it must be released/approved
+for use via a `TemplatePackage` and then assigned to a study. The workflow for
+this would be:
+
+1. Create a template package
+    - Add name, description
+    - Add templates
+    - Assign version number
+2. Assign the template package to a study
```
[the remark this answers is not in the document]
znatty22: Yea, sorry that was a mistake. I updated it to `studies`

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+
+Contains the submission rules and a reference to the actual submission template
+file on a storage system (e.g. local, S3).
+
+#### Database Table Schema
+
+| Name          | Type        | Description                                                                                                                                                                                        |
+| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| id            | varchar     | A human friendly identifier (Kids First style) for the submission template                                                                                                                         |
+| data_type     | char        | Short code identifying the type of data described by the template (e.g. PT_DIS for participant diseases)                                                                                           |
```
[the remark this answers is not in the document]
znatty22: I'm really not sure if we need `data_type` anymore? They would be determined by the author of the template. Its sort of analogous to a study short code.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+    - Add templates
+    - Assign version number
+2. Assign the template package to a study
+
+### CreateTemplatePackageMutation
+
+Creates a TemplatePackage and optionally assigns it to a study
+
+```graphql
+type CreateTemplatePackageInput {
```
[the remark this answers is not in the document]
znatty22: Hmm I wasn't sure if we should have an organization tied to the template package as well. My idea was that an organization owns and authors a template. And then someone can create a template package with templates that could come from more than one org. 

I wonder what would the purpose of a foreign key to org be on the package? Would that imply that there are additional restrictions to updating the package?

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+
+Updates a TemplatePackage and optionally assigns it to a study.
+
+```graphql
+type UpdateTemplatePackageInput {
+    id: ID!
+    name: String
+    description: String
+    submissionTemplates: [ID]
+    version: String
```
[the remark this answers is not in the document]
znatty22: Idk if it even makes sense to have a version on the template package (a.k.a a templates release) if we don't support template versioning. Let's talk about whether we want to support template versioning?

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
dankolbman: So then there would be a super directory for any user/admin in any org to browse through and try to incorporate into one of their studies? There's a lot of use cases to unwrap with that.

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
dankolbman: I think it can be removed.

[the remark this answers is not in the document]
dankolbman: Data type really is being replaced by the template concept.

docs/rfcs/0003_templates.md
znatty22: Let's talk about whether we want to support template versioning

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
znatty22: Yea probably right. Ok so a template package also has an owning org like templates do. And then the update/delete ops for packages will be allowed for org admins that are from the owning org only. No one from outside the org can update/delete packages.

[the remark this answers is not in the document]
dankolbman: Do you think it may be similar to how Cavatica apps work? Maybe easier to 'clone' them into an org?
This sort of thing is easy to implement but I'm afraid there's so many edge cases we are gonna hit.

- Content moderation (make sure things have good names/descriptions)
- Allow things to be sharable/private
- Build out directory/template browser
- Deal with things like the owning org removing a template other's use
- People wanting to take a template and modify it for their needs

Those are just a few I can think of right now.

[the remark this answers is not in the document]
znatty22: Ok, I updated the MVP. I think this should work decently for an MVP. It alleviates some of those ^ issues via template versions. The longer term solution (if we choose to implement) incorporates package versioning which just makes it easier to manage templates in a logically related group and provides an avenue for a release process.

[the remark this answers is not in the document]
znatty22: I also wrote up a use case and solutions doc that outlines some use cases and walks through potential solutions for the most common one (modify a template that's being used by others)

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+    templates in the Study's template versions.
+-   The study and file mutations should be updated accordingly as well
+
+## 💻 Implementation Considerations
+
+When thinking about implementing data submission templates, we should consider
+the entire development, test, and documentation lifecycle and some of the
+common painpoints encountered as they may inform a better implementation of our
+templates.
+
```
dankolbman, line 320: We should think through how 'migrating' to a new template version happens.
We can create a new version of a template based on the latest version of a template, but how do we 'migrate' the files that use that template?
This could be easy if the change does something like remove a field. In this case, we can go through each document that used the old version and change it to the new one. This is possible because the document is valid as either the old or the new version.

However, it's a problem if we want to do something like add a field in a new version. All documents that use the old version will not be valid for the new version. This means we will have to add the new field to all the documents and upload new versions of them before we can change the document's template to the latest version.

So updating a document to match a new template version is possible, but it may not have the best user flow.
Something else to watch for is during this transition, some documents may use an old version while some may use a new version, but all documents will still be said to use the 'such and such template. This could be misleading unless we label documents with 'not latest version' badges or similar when browsing through documents and the templates they use.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+A change in my template definition (e.g. new column) should not cause a
+cascading set of manual changes that need to be made in several other
+related artifacts.
+
+For example, if I add a new column to the Participant Diseases template,
+I need to ensure that the following are also updated to reflect the change:
+
+-   Template documentation / end-user guidance
+-   Test data generators in template unit tests
```
dankolbman, line 331: If we are updating our tests to account for all real-world templates, it's not a unit test. Unit testing would be around features of templates where we would use templates that clearly demonstrate desired behavior with the minimal amount of definition.

docs/rfcs/templates_solutions.md
```diff
@@ -0,0 +1,314 @@
...
+
+3. Copy `BCM` -> `BCM_V1`, Update `BCM_V1` with desired changes
+
+    ```
+    Templates
+    ---------
+    PDE, BCM, BCM_V1
+    ```
+
+4. Replace `BCM` with `BCM_V1` in CBTN
```
dankolbman, line 70: what about all the documents in CBTN that are still using `BCM`?

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+| modified_at   | varchar     | Timestamp of update                                                                                                                                                                                |
+| creator       | foreign key | Foreign key to User that created the submission template                                                                                                                                           |
+
+#### Django Model Schema
+
+| Name                          | Type           | Description                                                                                            |
+| ----------------------------- | -------------- | ------------------------------------------------------------------------------------------------------ |
+| All fields in DB table schema |
+| released                      | BooleanField   | A computed field (not stored in db) denoting whether the template is being used in one or more studies |
+| submission_template_file      | FileField      | Reference to the actual template file in a storage system                                              |
```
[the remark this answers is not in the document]
znatty22: Yea, I was thinking the submission template would be computed from the fields schema. It might look something like this: 
image

docs/rfcs/templates_solutions.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Pros/Cons
+
+-   ✅ Templates are easily shareable among studies
+-   ✅ Can reliably/programmatically track provenance of templates
+-   ✅ Studies are not affected by non-relevant template updates
+-   ✅ Easier to manage a set of logically related templates and add to multiple studies
+-   ✅ Easier to track / capture a batch of template updates using template package versions
+-   ⚠️ UI design and frontend implementation becomes considerably more complex
+-   ⚠️ Backend implementation becomes considerably more complex
```
dankolbman, line 314: Recall that the requirement for schema enforcement on uploaded files comes from the need to be able to shovel this data into downstream pipelines. Considering that, it seems like each organization will naturally have one or two 'base' models that each study will need to adhere to. This seems to work pretty well with the package concept. However, there is a further need to have the supplemental fields, perhaps per-study. So each study will match up with the latest (hopefully) package version, but may have a handful of 'supplemental' templates that are specific to that study or the requirements of that study.

Perhaps we could consider these supplemental templates for the future. Going this way would probably require one package per study with the goal being that all studies really only target the latest version of that package. If we go with package versions, maybe we should consider restricting each study to only implement one package at a time.

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
dankolbman: In this case, we shouldn't need to store anything in the database since we can generate it on demand.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+referenced template version and it's template. Overwrites the name,
+description, and owner of the source Template with the supplied values
+in the cloned Template.
+
+```graphql
+type CloneTemplateInput {
+    templateVersion: ID!
+    name: String!
+    description: String!
+    owner: ID
```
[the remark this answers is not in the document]
znatty22: Yes. I can change it to organization to make it more clear

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+### (Optional) CloneTemplateMutation
+
+Creates a new Template and a TemplateVersion by duplicating the
+referenced template version and it's template. Overwrites the name,
+description, and owner of the source Template with the supplied values
+in the cloned Template.
+
+```graphql
+type CloneTemplateInput {
+    templateVersion: ID!
```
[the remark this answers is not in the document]
znatty22: What do you mean? I didn't include it in the mockups, but I'm assuming whatever UI we build for creating a new template version would allow you to create via copying an existing version.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+## ✨ Creating Templates
+
+### CreateTemplateMutation
+
+Creates a Template and initial TemplateVersion
+
+```graphql
+type CreateTemplateInput {
+    name: String!
+    owner: ID!
```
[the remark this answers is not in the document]
znatty22: yea

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+    templateVersion: ID!
+    name: String!
+    description: String!
+    owner: ID
+}
+```
+
+### Permissions
+
+-   Any organization or study admin may create or clone Templates/TemplateVersions
```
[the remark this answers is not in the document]
znatty22: Ok, wasn't sure. I can remove that bit

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+
+-   Any organization or study administrator may update a Template or
+    TemplateVersion if their organization owns it
+
+## 🗑 Deleting Templates
+
+### DeleteTemplateMutation
+
+```graphql
+type DeleteTemplateInput {
```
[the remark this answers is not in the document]
znatty22: See Template Management Rules - you would not be able to delete the template if studies are using any of the versions in this template.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+
+Creates a Template and initial TemplateVersion
+
+```graphql
+type CreateTemplateInput {
+    name: String!
+    owner: ID!
+    description: String!
+    fieldsSchema: JSONString
+    fieldsSchemaFile: Upload
```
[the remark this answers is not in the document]
dankolbman: We can parse it and post it to the url from the front end, no need to deal with both from the api.

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
dankolbman: Are these versions found through browsing through all template versions in the current organization or is there some public template listing?

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Download Study Templates
+
+Probably the easiest way to download templates since it bundles everything
+into an archive. You will get an archive containing:
+
+For each TemplateVersion in the study:
+
+-   TemplateVersion.submission_template_file: the TSV file that people will fill out and use for data submission to their study
+-   TemplateVersion.fields_schema: the template field definitions which serve as instructions on filling out the template
```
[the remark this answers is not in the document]
znatty22: This line is misleading. You wouldn't actually download the `fields_schema` JSON. Ideally, you'd get a multi-sheet Excel workbook where each sheet has the field definitions for a template. OR if we don't want to assume people have Excel, you'd get a TSV file of field definitions for each template. The Excel workbook approach would look something like this:

image

I think the field definitions/schema is the most important part of template distribution

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+Download the latest version of template.
+
+```
+/download/template/<Template.id>
+```
+
+You would get an archive with two files:
+
+-   Template.latest_version.submission_template_file
+-   Template.latest_version.fields_schema
```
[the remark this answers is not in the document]
znatty22: See earlier comment about `fields_schema`. I think that's the most important part of the "template" because it defines each field, provides type, instructions, etc for the actual submission template

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+### Permissions
+
+### Permissions
+
+-   Any organization or study administrator may delete a Template or
+    TemplateVersion if their organization owns it
+
+## 📦 Downloading Templates
+
+### Download Study Templates
```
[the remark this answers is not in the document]
znatty22: Yes I think so

docs/rfcs/templates_solutions.md
znatty22: I haven't covered data migration when template definitions change. Still have to think about that

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+-   Any user may download template files
+
+## Changes to Existing Functionality
+
+-   A `Study` will have a new field, `Study.template_versions`, a many to many
+    relationship to `TemplateVersion`.
+-   The `File.file_type` field will be replaced by a one to one foreign key
+    to a `TemplateVersion`. The `TemplateVersion` must be one of
+    templates in the Study's template versions.
+-   The study and file mutations should be updated accordingly as well
```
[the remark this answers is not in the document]
dankolbman: Right now we take all the column names and see which data types are subsets of them. This way we can tell the user which types are valid selections for the file.

docs/rfcs/templates_solutions.md
znatty22: Yea I agree - a Study has one PackageVersion (aka foreign key) at a time. I thought that was represented in the model diagram. If not, that's a mistake.

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
znatty22: Ok, wasn't sure. I'll update the doc

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
znatty22: The template list view in the PR description _is_ the public template listing. You can filter by study (and I think I forgot to add a filter for organization). Now that list view has rows that represent the `Template` not `TemplateVersion`. So if we added a copy/clone button the actions in the row, clicking it would copy the latest version of the `Template` and create a new `Template` and `TemplateVersion`. 

I haven't thought enough about the UI for copying a specific version of a template to create a new `Template` + `TemplateVersion`. But I imagine that could happen via the Edit Template View. Could add a copy button to the actions in the Versions list view, and then prompt the user to create a new TemplateVersion OR an entirely new template (Template + TemplateVersion).

Also, keep in mind that even if we didn't support creation through cloning, someone could always download a template (field defs, template file, etc) and manually create a new Template + TemplateVersion.

docs/rfcs/0002_templates_mvp.md
znatty22: Yea, I think migration deserves a separate RFC or addendum. Haven't thought enough about it yet

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
znatty22: I don't know if that's a good idea. Could be but we might need to revise the user flow. Right now it's difficult to tell why a file type isn't showing up in the valid types list. 

For example, I upload a file that has 50 columns in it and it matches with lots of file types/templates: participant details, participant diseases, participant outcomes, participant ..., and specimen collection manifest. But for some reason the valid file types doesn't include the biobank manifest. Why? Ah its because I'm missing one column (e.g. Aliquot ID).

It might be better to allow the user to select the template they think their file conforms to. And then tell them if their file is missing required columns.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+You would get an archive with two files:
+
+-   TemplateVersion.submission_template_file
+-   TemplateVersion.fields_schema
+
+### Permissions
+
+-   Any user may download template files
```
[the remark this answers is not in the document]
znatty22: I think any user should be allowed to download a template. It gives people a peek at what kind of data the organization is trying to collect. So anyone curious should be able to see that

docs/rfcs/0002_templates_mvp.md
znatty22: Yea I agree. I just think it would be nice if we could define the templates and then somehow generate data that conforms to the fields schema automatically so that it could easily be used in testing. It will also make testing extract configs easier (more on this later).

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
+# 🗃 Data Submission Templates
+
+**Status**: Approved, waiting for implementation
+
+## 💡 Motivation
+
+As the Kids First DRC handles a wide variety of disease types and research studies,
+we have been fairly flexible over the years in accepting a wide variety of formats
```
gsantia, line 8:
```suggestion
we have been fairly flexible over the years accepting a plethora of formats
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+with fairly loose requirements on the submission for clinical data.
+
+This approach made it easy for investigators whose studies already completed
```
gsantia, line 11:
```suggestion
This approach made it easy for investigators whose studies had already completed
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+to retroactively submit their data to the DRC. However, it would often make
+the clinical data validation and review process quite manual and drawn out
+since every study had a different set of files and formats that had to be
```
gsantia, line 14:
```suggestion
to retroactively submit their data to the DRC. However, it often
elongated the clinical data validation and review process with manual
work; every study had a different set of files and formats that had to be
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+interpreted, cleaned, and validated.
+
+This experience has motivated a submission mechanism that is standardized
+across studies and enables automated data validation.
+
+## 📝 Definition
+
+A data submission template is term that describes two things:
```
gsantia, line 22:
```suggestion
The term data submission template describes two things:
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+
+Each template typically describes submission standards for a particular entity
+(e.g. participant disease event) and each row in the tabular submission file
+represents one instance of that entity (e.g. cleft lip observed for PT1 at 400 days since birth).
+
+## 👪 Sharing
+
+Templates should be easy to reuse across organizations/studies.
+This is important because oftentimes organizations/studies will have such similar data
+collection requirements to each other that they may want to reuse existing templates
```
gsantia, line 39:
```suggestion
collection requirements that they may want to reuse existing templates
```

docs/rfcs/0002_templates_mvp.md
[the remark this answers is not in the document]
gsantia: Could also let them know if there are other templates that their file _does_ conform to

docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,353 @@
...
+#### Database Table Schema
+
+| Name         | Type        | Description                                                                                              |
+| ------------ | ----------- | -------------------------------------------------------------------------------------------------------- |
+| id           | varchar     | A human friendly identifier (Kids First style) for the template                                          |
+| data_type    | char        | Short code identifying the type of data described by the template (e.g. PT_DIS for participant diseases) |
+| name         | varchar     | Short name for template                                                                                  |
+| description  | text        | A short blurb describing the entity the template represents                                              |
+| created_at   | datetime    | Timestamp of creation                                                                                    |
+| modified_at  | varchar     | Timestamp of update                                                                                      |
```
gsantia:
```suggestion
| modified_at  | datetime     | Timestamp of update                                                                                      |
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+| name          | varchar     | Short name for submission template                                                                                                                                                                 |
+| description   | text        | A short blurb describing the entity the submission template represents                                                                                                                             |
+| fields_schema | JSON        | The template submission rules. A JSON blob with a standard structure which will encode the template field objects having things like: label, key, description, type, values_accepted, instructions |
+| created_at    | datetime    | Timestamp of creation                                                                                                                                                                              |
+| modified_at   | varchar     | Timestamp of update                                                                                                                                                                                |
```
gsantia:
```suggestion
| modified_at   | datetime     | Timestamp of update 
```

docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,353 @@
...
+| id            | varchar     | A human friendly identifier (Kids First style) for the template version                                                                                                                            |
+| name          | varchar     | Short name for template version                                                                                                                                                                    |
+| description   | text        | A short blurb describing the version (e.g. changes made since last version)                                                                                                                        |
+| created_at    | datetime    | Timestamp of creation                                                                                                                                                                              |
+| modified_at   | varchar     | Timestamp of update                                                                                                                                                                                |
```
gsantia:
```suggestion
| modified_at   | datetime     | Timestamp of update
```

docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,353 @@
...
+
+## ✨ Creating Templates
+
+Templates can be created from scratch or via copy of an existing template
+
+For both the create mutations:
+
+-   If `fieldsSchema` is supplied then use that to populate
+    `TemplateVersion.fields_schema`.
+-   Otherwise, if `fieldsSchemaFile` it is supplied then parse it into
```
gsantia:
```suggestion
-   Otherwise, if `fieldsSchemaFile` is supplied then parse it into
```

docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Permissions
+
+-   Any organization or study administrator may create a template
+    package.
+-   Any organization or study administrator may assign a
+    template package to a study that they have access to.
+-   Any organization or study administrator may update a template
+    package.
+-   Any organization or study administrator may delete a template
```
gsantia: Will they be able to create template packages they didn't create?

znatty22: I think you're looking at an old version of the RFC? I got rid of the `0003_templates.md` doc

fiendish:
I think this can/should be unified with https://github.com/kids-first/kf-api-study-creator/pull/571

For instance, I don't see any way to establish the sets of acceptable value types and constants as shown in 



in this format definition screenshot segment: 
Screen Shot 2021-06-01 at 1 44 02 PM

The json object could/should include e.g. `{"accepted_values": ["Male", "Female", "Other"]}` or `{"accepted_values": [hipaa_age_class]}` etc.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,372 @@
...
+
+#### Database Table Schema
+
+| Name          | Type        | Description                                                                                                                                                                                        |
+| ------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| id            | varchar     | A human friendly identifier (Kids First style) for the submission template                                                                                                                         |
+| description   | text        | A short blurb describing this version of the template                                                                                                                                              |
+| fields_schema | JSON        | The template submission rules. A JSON blob with a standard structure which will encode the template field objects having things like: label, key, description, type, values_accepted, instructions |
```
shanewilson, line 139: Is this expanded on somewhere? It seems like this is the functional bit of the template - is there a document that goes over the different rules? how they are read, how dynamic they are, if they can access row / column / table data etc.

znatty22: Hey @shanewilson, first, sorry for the long response! It might be worth it to have a chat about this later. I'm sure this response doesn't answer all your questions

So there isn't a separate document to describe the `fields_schema` at the moment. In this first iteration, the field definitions will be very simple (label/column name, key/canonical attribute, column type, and values accepted). Later on we may expand on this and allow field definitions to include some sort of validation functions (e.g. age range check). 

Maybe some background would be useful. Currently, most of the heavy lift data validation is done after the submitted data file has been mapped (manually done by an analyst) to a set of standard column names understood by the study creator's validation library. It checks for things like: participants with more than one gender value, specimens that are not linked to any participant, genomic files missing associated sequencing experiment information, etc.

The rough workflow todayish/near future is:

1. Analyst uploads a tabular file to Data Tracker which conforms to one of the Kids First data templates (See [this PR](https://github.com/kids-first/kf-api-study-creator/pull/571) for details)
2. Analyst creates a data review on Data Tracker
3. Analyst adds their files to the data review
4. Analyst pushes a button to run validation
5. Validation is able to run on any data files that conform to the Kids First templates

As you might imagine the current process ^ needs to change in order make the validation a bit less hard coded. This is where defining validators for each field might help us. This is pretty big topic though and will require a separate RFC. It might help if you [take a peek](https://kids-first.github.io/kf-lib-data-ingest/tutorial/testing.html#data-validation) at how our analysts use the Kids First ingest/validation library to prepare the data and validate. I'm sorry that documentation is so meager but it might help.

When you say `different rules` - are you talking about field validation?

Can you expand on what you mean by `how dynamic they are` and what you mean by `how they are read`? 

I wonder if it would help to show you the json schema / structure I'm thinking of right now for the `fields_schema`. I figured I would expand on this later on when I get close to implementing the template versions.

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+| extract_config_file           | FileField      | Reference to the actual extract config file in a storage system                                                      |
+| template_packages             | RelatedManager | List of template packages this template is contained in                                                              |
+
+### TemplatePackage
+
+A template package is a set of submission templates that will be used to set
+submission standards and collect data for a study.
+
+The concept of a template package is useful for several reasons. First,
+templates are typically used in conjunction with one and other because a set of
```
gsantia:
```suggestion
templates are typically used in conjunction with one another because a set of
```

docs/rfcs/0002_templates_mvp.md
```diff
@@ -0,0 +1,359 @@
...
+templates usually represents one data model, and each template typically
+captures data for a single entity in the model.
```
gsantia: I'm not sure what to suggest here because I don't really understand what you're trying to say. I assume `with one and other` is supposed to be `with one another`, but I'm still not entirely sure what you mean.

docs/rfcs/0003_templates.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+### UpdateTemplateVersionMutation
+
+Updates a TemplateVersion. A TemplateVersion may be updated if it is not used
+in a study (part of a TemplatePackage assigned to a study)
+
+-   If `fieldsSchema` is supplied then use that to populate
+    `TemplateVersion.fields_schema`.
+-   Otherwise, if `fieldsSchemaFile` it is supplied then parse it into
```
gsantia:
```suggestion
-   Otherwise, if `fieldsSchemaFile` is supplied then parse it into
```

docs/rfcs/0002_templates_mvp.md
shanewilson: Thanks for the response @znatty22 

> When you say different rules - are you talking about field validation?

yea I'm asking about built-in rules I guess. Like can the template say a column is OneOf("A","B","C") and is there a list of those built-in rules?

> Can you expand on what you mean by how dynamic they are and what you mean by how they are read?

by dynamic I mean can the rules reference other bits of data? Like instead of OneOf(A,B) it could be OneOf(column.fieldA.values) or whatever

by how are they read I'm really asking about text vs code - how does a rule get defined and how is that turned into code. Is the rule just text? and then there is some text parsing to figure out which validation function to run?
I know in the ICGC submission they evenually added in https://en.wikipedia.org/wiki/MVEL (I think) so rules could be fully dynamic. It was super terrible because you basically had to write code in a text box. Just wondering what the plan is for when we have to start handling more complicated and dynamic rules.

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+
+![Field Definitions](../_static/images/template-field-defs.png)
+
+## High Level User Flow
+
+TODO: Add diagram?
+
+### 1. Build Template
+
+-   Admin: Build template on Data Tracker (e.g. define the template fields)
```
gsantia, line 37: How is a template built on Data Tracker? Is there an interface for that?

znatty22: Yea and its TBD right now. I may try an mock up something this week for it

znatty22: For the MVP, building the template may just mean uploading the field definitions in a TSV file or authoring the field definitions JSON in some sort of input field

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+focusing on validation of a single submitted data file at a time.
+
+A user should be able to upload a data file, select which template they think
+it conforms to (or it could get auto-assigned based on column presence), and
+then click a button to validate the file for the following:
+
+1. Are all required columns present?
+2. Does every record in a required column have a non-empty value?
+3. For any columns that have an `acceptable_values` restriction, do all values
+   take on a value from that list
```
gsantia, line 258:
```suggestion
    take on a value from that list?
```

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+
+1. Analyst: Download Extract Config for a template
+1. Analyst: Edit it to include additional wrangling logic
+1. Upload Extract Config for the template
+
+**Submit Data**
+
+1. Data Submitter: Upload data file claiming to conform to a template
+2. API: Perform basic single file validation
+3. Data Submitter: Receive immediate validation feedback on data file (e.g. missin required column "Gender")
```
gsantia, line 332:
```suggestion
3. Data Submitter: Receive immediate validation feedback on data file (e.g. missing required column "Gender")
```

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+4. Data Submitter: Fix any errors in file, repeat steps 1-3
+
+**Review and Validate Submission**
+
+1. Analyst: Create data review and add data file to the review
```
gsantia, line 337:
```suggestion
1. Analyst: Create data review and add data files to the review
```

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+6. Analyst: View validation report on Data Tracker and take actions
+
+The `Review and Validation Submission` features is actually already implemented
+in Data Tracker and Study Creator API, thus this approach is very feasible for an MVP
+solution.
+
+## API Details
+
+```python
+
```
gsantia, line 352: where's this?
`creator/data_templates/models.py` ?

docs/rfcs/0003_template_fields_mvp.md
```diff
@@ -0,0 +1,406 @@
...
+def download_study_templates(request, study_id):
+    """
+    For all templates assigned to a study, generate template field definitions
+    file and submission files for each template
+    Zip up above files into archive
+    Return archive content in HTTP response
+    """
+    pass
+```
```
gsantia, line 406: Maybe a user would want to download all templates assigned to a study that their organization created?

docs/rfcs/0003_template_fields_mvp.md
gsantia: Cool I wasn't sure if I was missing something in this doc

docs/rfcs/0002_templates_mvp.md
znatty22: @shanewilson Hopefully these [slides](https://docs.google.com/presentation/d/1Qb425xdqGt0VKlCO81edyCzlCqohm94J1QwxxGd2ajY/edit?usp=sharing) and [this write up](https://github.com/kids-first/kf-api-study-creator/blob/templates-rfc/docs/rfcs/0003_template_fields_mvp.md) will answer some of your questions!

znatty22: Relating to your last point on en/decoding validation logic - I haven't searched around much on the topic but I remember when I started doing this before, I had come across CUE - an open-source data validation language: https://cuelang.org/docs/about/. It looks kinda confusing tho lol. Anywho, I will think about this more deeply when we get closer to designing the next version of validation in Dewrangle

PR #655, opened 2021-05-27
♻️ Add study attribute to IngestProcess

### LINKED ISSUE #654: IngestProcess should have a study attribute
The `IngestRun` class (a child of `IngestProcess`) has a `study` attribute which is not strictly mandatory. Instead, we should put this attribute in the parent class and make sure it throws a `NotImplementedError` if it's accessed.

gsantia:
Closes #654 

Gives the parent class `IngestProcess` a `study` attribute which yields `NotImplementedError` when accessed, and adds a test to check this. Also applies black formatting to the relevant files.

### DIFF
--- a/creator/ingest_runs/common/model.py
+++ b/creator/ingest_runs/common/model.py
@@ -109,6 +109,13 @@ class IngestProcess(models.Model):
         """
         return django_rq.get_queue(name=self.__queue__)
 
+    @property
+    def study(self):
+        """
+        This should not be accessed.
+        """
+        raise NotImplementedError
+
     @transition(field=state, source=State.NOT_STARTED, target=State.RUNNING)
     def start(self):
         """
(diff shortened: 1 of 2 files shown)


creator/ingest_runs/common/model.py
```diff
@@ -89,6 +94,7 @@ class Meta:
         default=State.NOT_STARTED,
         help_text="The current state of the ingest process"
     )
+    study = AbstractAttribute()
```
dankolbman, line 97: Hm any reason in particular you chose to create a new attribute class with a custom getter?

The way I would usually see this done is via the property decorator on a class method:
```python
class IngestProcess(model.model):
    @property
    def study(self):
        raise NotImplementedError
```

gsantia: > Hm any reason in particular you chose to create a new attribute class with a custom getter?
> 
> The way I would usually see this done is via the property decorator on a class method:
> 
> ```python
> class IngestProcess(model.model):
>     @property
>     def study(self):
>         raise NotImplementedError
> ```

I'm not entirely sure. I was following this:
https://stackoverflow.com/a/32536493

Seems the property works the same here

dankolbman (approved):
# :garlic:

PR #660, opened 2021-06-09
♻️ Overwrite S3 files when appending

dankolbman:
Job logs currently accumulate as Django appends strings to the file names if the file already exists. Job logs are intended to be appended to, so this behavior is unnecessary. We do, however, need to make sure versions do not get overwritten by other versions that share the same name so we need to prepend a uuid.

logs

### DIFF
--- a/creator/settings/production.py
+++ b/creator/settings/production.py
@@ -338,6 +338,7 @@ UPLOAD_DIR = os.environ.get("UPLOAD_DIR", "uploads/")
 
 # This will be dynamically set depending on the file being requested
 AWS_S3_BUCKET_NAME = "kf-study-us-east-1-dev-sd-me0owme0w"
+AWS_S3_FILE_OVERWRITE = True
 
 # Auth0 settings
 AUTH0_DOMAIN = os.environ.get("AUTH0_DOMAIN", "https://kids-first.auth0.com")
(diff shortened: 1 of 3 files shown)


gsantia (approved):
☕️

creator/settings/production.py
```diff
@@ -338,6 +338,7 @@ def traces_sampler(sampling_context):
 
 # This will be dynamically set depending on the file being requested
 AWS_S3_BUCKET_NAME = "kf-study-us-east-1-dev-sd-me0owme0w"
+AWS_S3_FILE_OVERWRITE = True
```
znatty22, line 341: I'm a bit confused..the docs say that the default for this setting is `True`. Does that mean that versions with the same name were getting overwritten in S3 before the change to prepend uuid?

znatty22: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html

dankolbman: Those are the docs for `django-storages`. We are using `django-s3-storage` where the default is `False`: https://github.com/etianen/django-s3-storage#file-storage-settings

znatty22 (approved):

gsantia:
Got a test failing now

gsantia (approved):

PR #662, opened 2021-06-09
♻️ Add timezone to Banner mutation datetimes

### LINKED ISSUE #659: Fix banner dates
We get these warnings when running tests:
```python
RuntimeWarning: DateTimeField Banner.end_date received a naive datetime (2021-06-10 14:49:21.632576) while time zone support is active.
```
```python
RuntimeWarning: DateTimeField Banner.start_date received a naive datetime (2021-06-09 14:49:21.632568) while time zone support is active.
```

This is because the datetimes being saved to the database do not have information about the timezone. We should add this info when saving dates to solve these warnings.

gsantia:
Closes #659 

Adds timezone information to the dates being saved in the Banner mutation and its tests.

### DIFF
--- a/creator/status/banners/mutations.py
+++ b/creator/status/banners/mutations.py
@@ -8,7 +8,7 @@ from creator.status.banners.models import Banner
 
 
 class BannerInput(graphene.InputObjectType):
-    """ Parameters used when creating a new banner """
+    """Parameters used when creating a new banner"""
 
     start_date = graphene.DateTime(
         required=False, description="When to start displaying the banner"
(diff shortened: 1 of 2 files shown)


tests/status/test_banner_mutations.py
```diff
@@ -73,15 +74,17 @@ def test_create_banner(db, clients, user_group, allowed):
                 "enabled": True,
                 "url": "https://www.example.com",
                 "urlLabel": "View more info here",
-                "startDate": datetime.datetime.now().isoformat(),
+                "startDate": timezone.now().isoformat(),
```
dankolbman, line 77: This fixes the warning in the tests, but what if the user sends in a date time without a timezone? Will it still warn? Maybe we should attach a default timezone?

gsantia: Good point. It would still warn. I'm going to need to figure out all the points where the user sends in datetimes.

dankolbman: This should be the only place as fair as I know. Usually datetimes are populated automatically.

creator/status/banners/mutations.py
```diff
@@ -1,13 +1,14 @@
 import graphene
 from graphql import GraphQLError
 from graphql_relay import from_global_id
+from django.utils.timezone import make_aware
```
dankolbman, line 4: Neat, didn't know about this.

dankolbman (approved):
# :clock1030:

PR #664, opened 2021-06-09
♻️ Fix pandas indexing warnings

### LINKED ISSUE #663: Fix pandas indexing warnings
The tests give us several warnings like this:

```
tests/ingest_runs/test_genomic_data_loader.py::test_load_harmonized_genomic_files
tests/ingest_runs/test_genomic_data_loader.py::test_load_specimen_harmonized_gf_links
tests/ingest_runs/test_genomic_data_loader.py::test_load_seq_exp_harmonized_genomic_files
  /Users/santiag/Projects/kf-api-study-creator/creator/ingest_runs/genomic_data_loader.py:204: SettingWithCopyWarning:
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead

  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df[CONCEPT.GENOMIC_FILE.HARMONIZED] = True
```

We should set the column values in the proper way

gsantia:
Closes #663 

When running tests we get a bunch of `SettingWithCopyWarning` notices from pandas.
`GenomicDataLoader` is refactored to get rid of these and avoid modifying copies of DataFrames.

### DIFF
--- a/creator/ingest_runs/genomic_data_loader.py
+++ b/creator/ingest_runs/genomic_data_loader.py
@@ -79,7 +79,7 @@ class GenomicDataLoader(object):
         logger.info(f"Getting file metadata from S3 {buckets} ...")
 
         file_df = utils.scrape_s3(buckets)
-        file_df = file_df[~file_df["Filename"].isnull()]
+        file_df.dropna(subset=["Filename"], inplace=True)
 
         genomic_df = manifest_df.merge(
             right=file_df, on="Filepath", how="inner"
@@ -114,7 +114,7 @@ class GenomicDataLoader(object):
                 CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID,
                 CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
             ]
-        ]
+        ].copy()
         biospec_gen_df[CONCEPT.BIOSPECIMEN.ID] = biospec_gen_df[
             CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID
         ]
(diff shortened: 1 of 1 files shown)


creator/ingest_runs/genomic_data_loader.py
```diff
@@ -115,10 +115,14 @@ def load_specimen_harmonized_gf_links(self, genomic_df):
                 CONCEPT.GENOMIC_FILE.TARGET_SERVICE_ID,
             ]
         ]
-        biospec_gen_df[CONCEPT.BIOSPECIMEN.ID] = biospec_gen_df[
-            CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID
-        ]
-        biospec_gen_df[CONCEPT.BIOSPECIMEN_GENOMIC_FILE.VISIBLE] = True
+        biospec_gen_df = biospec_gen_df.assign(
```
dankolbman, line 118: If I'm reading this correctly, you are simply renaming this column? Could you use `rename` instead?

gsantia: No, I'm creating a new `CONCEPT.BIOSPECIMEN.ID` column with the same values as the `CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID` column. The ingest library requires both

dankolbman: I see, in that case, does `df.loc[:, CONCEPT.BIOSPECIMEN.ID] = df['CONCEPT.BIOSPECIMEN.TARGET_SERVICE_ID]` work? It seems a lot less convoluted.

creator/ingest_runs/genomic_data_loader.py
```diff
@@ -199,10 +205,15 @@ def _prep_harmonized_gf_df(self, df):
...
             "Source Read": CONCEPT.GENOMIC_FILE.SOURCE_FILE,
         }
-        df = df[list(col_rename.keys())]
-        df.rename(columns=col_rename, inplace=True)
-        df[CONCEPT.GENOMIC_FILE.HARMONIZED] = True
-        df[CONCEPT.GENOMIC_FILE.VISIBLE] = True
+
+        df = df[list(col_rename.keys())].rename(columns=col_rename)
+
+        df = df.assign(
```
dankolbman, line 211: Does using `df.loc` work here?

creator/ingest_runs/genomic_data_loader.py
gsantia: Yeah you're right

gsantia: That still gives the same warning

dankolbman (approved):
# :bear:

PR #665, opened 2021-06-10
⬆️ Bump Django version to fix regression

dankolbman:
Bump Django version to deal with regression regarding path elements in file names.


znatty22 (approved):
👍

gsantia (approved):

gsantia:
lots of tests failing?

PR #667, opened 2021-06-10
♻️ Address graphene and markdown warnings

### LINKED ISSUE #648: Address warnings
Grahpene and markdown give the following warnings that we should address to silence them:
```
/usr/local/lib/python3.7/site-packages/graphene/utils/subclass_with_meta.py:52
  /usr/local/lib/python3.7/site-packages/graphene/utils/subclass_with_meta.py:52: PendingDeprecationWarning: Defining `exclude_fields` is deprecated in favour of `exclude`.
    super_class.__init_subclass_with_meta__(**options)
```

```
/usr/local/lib/python3.7/site-packages/graphene_django/types.py:118
  /usr/local/lib/python3.7/site-packages/graphene_django/types.py:118: UserWarning: Excluding the custom field "file_type" on DjangoObjectType "FileNode" has no effect. Either remove the custom field or remove the field from the "exclude" list.
    type_=type_,
```

```
/usr/local/lib/python3.7/site-packages/markdown/util.py:87
  /usr/local/lib/python3.7/site-packages/markdown/util.py:87: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    INSTALLED_EXTENSIONS = metadata.entry_points().get('markdown.extensions', ())
```

gsantia:
Closes #648 .

Silences warnings made by graphene_django and markdown during testing. For markdown, updating to Python 3.8 is required.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
       ignore:
         - gh-pages
     docker:
-      - image: python:3.7
+      - image: python:3.8
         environment:
           - PG_USER=study-creator
           - PG_PASS=password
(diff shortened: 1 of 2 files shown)


gsantia:
This doesn't address the markdown warning found in the issue because I don't get that when I run the tests and am unsure where it comes from.

Also, it seems like all we need to do is delete this one line. All the tests pass. What was it for? Perhaps we're using a newer version of `graphene_django` now and it isn't necessary any more?

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       ignore:
         - gh-pages
     docker:
-      - image: python:3.7
+      - image: python:3.9
```
znatty22, line 9: This should probably match the version in our Dockerfile so that our CI and deployments are using the same Python version. I _think_ we're on 3.8?

gsantia: `FROM python:3.8-slim-buster as base`

ok, i'll try that!

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       ignore:
         - gh-pages
     docker:
-      - image: python:3.7
+      - image: python:3.8
```
znatty22, line 9: Are the changes in the circleci config needed to get rid of the warnings?

gsantia: The markdown warning only happens when using python 3.7 or lower, as far as I can tell

dankolbman (approved):
Looks good to me. It seems like the enum issue that was there with older versions of graphene-django isn't a problem anymore.

znatty22 (approved):

PR #670, opened 2021-06-14
⬆️  Bump pytest and ingest library versions

### LINKED ISSUE #668: Update pytest
It would be handy if we could update to the latest pytest. This may also help with conflicts that we have in the ingest library.

gsantia:
Closes #668 .

Updates pytest to the latest version and also updates the pytest plugins we're using. Makes the necessary modifications for these updates to occur. Silence new warnings that appear except for those caused by graphene, boto3, graphene_django, and markdown. The graphene_django and markdown warnings will be taken care of in another PR. 

Update to the latest release of the ingest library, as this has the flexible pytest requirement. Make small changes to facilitate this bump.


dankolbman (approved):
# :pager:

znatty22 (approved):
Looks ok, but at some point we should do a real test for the GWO ingest stuff

PR #672, opened 2021-06-17
✨ Add node query for querying nodes

shanewilson:
Add a top level `node` query that can query for and return any node type. Looks like it is built-in to graphene just had to add it to the schema.

```
query {
  node(id: "RmlsZU5vZGU6U0ZfMDAwMDAwMDA=") {
    ... on FileNode {
      id
      name
    }
  }
}
```

```
{
  "data": {
    "node": {
      "id": "RmlsZU5vZGU6U0ZfMDAwMDAwMDA=",
      "name": "assume.avi"
    }
}
```

### DIFF
--- a/creator/schema.py
+++ b/creator/schema.py
@@ -40,6 +40,7 @@ class Query(
 ):
     """ Root query schema combining all apps' schemas """
 
+    node = graphene.relay.Node.Field()
     if settings.DEBUG:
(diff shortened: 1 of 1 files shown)


dankolbman:
Force-pushed a linting fix.

dankolbman (approved):
# :1st_place_medal:

znatty22 (approved):

PR #680, opened 2021-06-18
🐛 Update Study model to auto-populate created_at

### LINKED ISSUE #673: Study created_at field can be None
Here's the attribute definition:
```python
created_at = models.DateTimeField(auto_now_add=False, null=True, help_text="Time the study was created")
```

`auto_now_add` defaults to `False`, so this isn't doing anything.

In the shell you can create a `Study` which will have no `created_at` value:

```
>>> from creator.studies.models import Study
>>> st = Study()
>>> st.save()
>>> print(st.created_at)
None
```

This doesn't come up in testing because the `Study` factory assigns a random time to this field.

gsantia:
Closes #673 

Changes the `created_at` attribute in the `Study` model to have `auto_now_add` be `True` instead of `False`.


gsantia:
Oops, forgot the migration!

dankolbman (commented):
This looks good. We probably did not automatically populate this column because originally we were using the Dataservice as the source of truth for studies. Now the study creator is the owner of studies so we should be good to set it here.

Can we wait until #632 merges as it has another study migration in it that includes a data migration so better not to mess with it.

znatty22 (approved):

PR #681, opened 2021-06-18
✨ Sort studies by last active

### LINKED ISSUE #520: Sort studies by Last Active
The `allStudies` query should have a parameter that allows `orderBy: "last_active"`.

This should sort and return studies based on the time of the last created version.

gsantia:
Closes #520 

Add a new parameter to the `allStudies` query to allow for sorting based on the time of the last created Version.

### DIFF
--- a/creator/studies/nodes.py
+++ b/creator/studies/nodes.py
@@ -4,6 +4,7 @@ from graphene_django import DjangoObjectType
 from graphene_django.filter import DjangoFilterConnectionField
 from creator.events.schema import EventNode, EventFilter
 from creator.files.schema.file import FileNode, FileFilter
+from creator.files.schema.version import VersionNode, VersionFilter
 from creator.releases.nodes import ReleaseNode
 from creator.releases.queries import ReleaseFilter
 from creator.studies.models import (
@@ -29,6 +30,12 @@ class StudyNode(DjangoObjectType):
         ReleaseNode,
         filterset_class=ReleaseFilter,
         description="List all releases for the study",
+    )
+
+    versions = DjangoFilterConnectionField(
+        VersionNode,
+        filterset_class=VersionFilter,
+        description="List all versions"
     )
 
     class Meta:
--- a/creator/studies/schema.py
+++ b/creator/studies/schema.py
@@ -33,7 +33,9 @@ logger.setLevel(logging.INFO)
 
 
 class StudyFilter(FilterSet):
-    order_by = OrderingFilter(fields=("name", "created_on", "organization"))
+    order_by = OrderingFilter(
+        fields=("name", "created_at", "organization", "versions__created_at")
+    )
 
     class Meta:
         model = Study
--- a/tests/studies/test_study_queries.py
+++ b/tests/studies/test_study_queries.py
@@ -1,0 +1,61 @@
+import pytest
+from graphql_relay import to_global_id
+
+from creator.studies.factories import StudyFactory
+from creator.files.models import Version
+
+LAST_ACTIVE_STUDIES = """
+query {
+    allStudies (orderBy: "-versions__createdAt") {
+     edges {
+       node {
+         id
+        kfId
+        name
+        createdAt
+       }
+     }
+  }
+}
+"""
+
+
+@pytest.mark.parametrize(
+    "user_group,allowed",
+    [
+        ("Administrators", True),
+        ("Services", False),
+        ("Developers", False),
+        ("Investigators", False),
+        ("Bioinformatics", False),
+        (None, False),
+    ],
+)
+def test_query_active_studies(db, clients, user_group, allowed):
+    """
+    Test that ordering the allStudies query by active_at functions as expected.
+    """
+    client = clients.get(user_group)
+    st = StudyFactory()
+    st.save
+    st2 = StudyFactory()
+    st2.save()
+    st3 = StudyFactory()
+    st3.save()
+
+    for study in (st, st2, st3, st):
+        Version(study=study, size=123).save()
+
+    st_id = to_global_id("StudyNode", st.kf_id)
+    st2_id = to_global_id("StudyNode", st2.kf_id)
+
+    resp = client.post(
+        "/graphql",
+        data={"query": LAST_ACTIVE_STUDIES},
+        content_type="application/json",
+    )
+    if allowed:
+        edges = resp.json()["data"]["allStudies"]["edges"]
+        assert len(edges) == 3
+        assert edges[0]["node"]["id"] == st_id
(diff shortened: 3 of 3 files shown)


creator/settings/development.py
```diff
@@ -68,6 +67,7 @@
     "creator",
     "corsheaders",
     'creator.ingest_runs.apps.IngestRunsConfig',
+    "creator.studies.apps.StudiesConfig",
```
dankolbman, line 70: You should be able to use `creator.studies` here.

dankolbman (commented):
This adds a field to a study that gets updated whenever the study gets saved. The issue here is that users want to be able to see sort studies by their last active date, which is usually determined by when a study last had a document or version uploaded. The documents are a different model with a foreign key to the study so when they get created, they don't actually update the study's row.

I believe this will need to be implemented using a custom ordering filter which sorts studies by the date of latest version created in each study. This won't be as easy because there is no materialized field on the study model for this and so the query to get this ordering is going to be a bit more complex. 
To do it this way, it looks like you need to override some of the internals of graphene like is being done in this example: https://stackoverflow.com/a/61543302
graphene has a handy way to write [custom filter sets](https://docs.graphene-python.org/projects/django/en/latest/filtering/#custom-filtersets), but annoyingly, it's not so easy with ordering filters.

If we want to keep the sorting simple, you could go with the materialized field like you have here and also consider updating it whenever a new version gets added to a study. This allows a simple order by query, but it will fragment the logic for how the last active date is determined. This is problematic because if we ever want to change that logic, maybe by also considering the last action in a review or a release for that study, we will need to plug in the same update snippet in each of the places where those things happen. This would open up the possibility of us easily losing track of all the possible ways a study can determine whether it was active or not.

dankolbman:
Before looking into how to do custom orderings, maybe try to see if adding something like this will work:
```
 order_by = OrderingFilter(
        fields=(
            ('versions__created_at',),
        )
    )
```

dankolbman:
Something else worth exploring may be using the custom ordering filter in django-filters, which is what graphene is using under the hood: https://django-filter.readthedocs.io/en/stable/ref/filters.html?highlight=ordering#orderingfilter

creator/settings/development.py
gsantia: The signal I created doesn't work using `creator.studies`

gsantia:
> If we want to keep the sorting simple, you could go with the materialized field like you have here and also consider updating it whenever a new version gets added to a study.

I'm confused. Is this different from what I did?

gsantia:
> Before looking into how to do custom orderings, maybe try to see if adding something like this will work:
> 
> ```
>  order_by = OrderingFilter(
>         fields=(
>             ('versions__created_at',),
>         )
>     )
> ```

`Study` has no `versions` field, right?

dankolbman:
Ah I see that is how you did it. I commented before those signals got pushed.

Do you think you could investigate using a more advanced sorting method rather than updating the row every time something happens that we want to consider an 'active' event? If it's easy, it may be preferable for the other stated reasons but if it's not, this approach will probably be ok.

gsantia:
> Before looking into how to do custom orderings, maybe try to see if adding something like this will work:
> 
> ```
>  order_by = OrderingFilter(
>         fields=(
>             ('versions__created_at',),
>         )
>     )
> ```

Doesn't this require a `versions` field on `Study`? Then I guess we'd need a `post_save` signal on `Version` that would add the versions as they get created to the appropriate `Study.versions`?

gsantia:
So I've been reading into all the links provided and elsewhere, and I've been unable to come up with a query that I'd consider easy. Maybe I'm missing something. But I think we can go with the method that I have here

gsantia:
There seems to be an issue with one of the migrations from the organizations update?

creator/settings/development.py
znatty22: @gsantia if you add something like this https://github.com/kids-first/kf-api-study-creator/blob/89176bd16e9fe2ea2cefdd8eb049b41b8742264e/creator/data_templates/__init__.py into your `apps/__init__.py` you can just have `creator.studies` in your settings files

znatty22:
@gsantia Curious to know if this works well, did you try https://github.com/kids-first/kf-api-study-creator/pull/681#issuecomment-865098170?

gsantia:
> @gsantia Curious to know if this works well, did you try [#681 (comment)](https://github.com/kids-first/kf-api-study-creator/pull/681#issuecomment-865098170)?

I'm a bit confused by this. Should I create a new `Study.versions` field and populate it?

znatty22:
Oh I'm sorry I forgot that studies don't have versions they have files..nvmind

dankolbman:
@znatty22 @gsantia I believe versions do have relations to studies these days.

gsantia:
> @znatty22 @gsantia I believe versions do have relations to studies these days.

You're right, there's a foreign key. Do you know how I should form the query?
I know that `created_at` in Graphene gets mapped to `createdAt`, but I don't know how to handle the `versions__` part.
I haven't been able to find any info online and have tried a bunch of permutations. This is what I mean:

this is the filter:
```
class StudyFilter(FilterSet):
    order_by = OrderingFilter(
        fields=(
            "name", "created_at", "organization", "active_at", "versions__created_at"
        )
    )

    class Meta:
        model = Study
        fields = ["name", "organization"]
```

and this is the GraphQL query:
```
query {
    allStudies (orderBy: "versions_createdAt") {
     edges {
       node {
         id
       }
     }
  }
}
```

Most of my attempts give:
```
"message": "['{\"order_by\": [{\"message\": \"Select a valid choice. versions_created_at is not one of the available choices.\", \"code\": \"invalid_choice\"}]}']"
```

gsantia:
> @znatty22 @gsantia I believe versions do have relations to studies these days.

Hey... It works!

creator/studies/nodes.py
```diff
@@ -31,6 +32,12 @@ class StudyNode(DjangoObjectType):
         description="List all releases for the study",
     )
 
+    versions = DjangoFilterConnectionField(
```
dankolbman, line 35: Why was this needed? Did it not allow you to sort by version properties without it?

dankolbman (commented):
I don't really like how obscure `orderBy: "-versions__createdAt"` is but the `orderBy` field is sort of naturally obscure anyway since it doesn't convert things to camel and doesn't really document what's allowed in the schema.

creator/studies/nodes.py
gsantia: yeah exactly

dankolbman (approved):
# 🧨

PR #687, opened 2021-06-29
♻️ Remove obsolete file types

### LINKED ISSUE #685: Remove obsolete file types
Now that we have the new file types which correspond to the KF templates, we should remove old file types that do not have templates and are too general to be meaningful. Those would be: `SHM - Shipping Manifest`, `CLN - Clinical Data`. We should keep the other two file types without templates: `OTH - Other`, `DBG - dbGaP file` because the `OTH` can be used for any file that doesn't match a template/expedited type and `DBG` may still be an important categorization. 

The frontend should also update its file types so that the old types are not displayed to the user.

## To do:

- [x] Data migration that will set `File.file_type=OTH` for any File that has a file type of `SHM`, `CLN`
- [x] Remove the old file types: `SHM`, `CLN` from `creator.analyses.file_types` 
- [x] Database schema migration to remove the old file type choices from `File.file_type`

gsantia:
Closes #685 

Remove `SHM` and `CLN` file types because they don't have templates and are too general to be useful. Set these values to `OTH`, and make the necessary code/database changes to reflect this.

- [x]  Data migration that will set `File.file_type=OTH` for any `File` that has a file type of `SHM`, `CLN`
- [x] Remove the old file types: `SHM`, `CLN` from `creator.analyses.file_types`
- [x] Database schema migration to remove the old file type choices from `File.file_type`


creator/files/migrations/0024_set_old_types_to_oth.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations
```
znatty22, line 1: Could you please combine the two migrations into one?

tests/files/test_valid_types.py
```diff
@@ -22,9 +22,7 @@ def test_default_types(db, clients, upload_file):
     file = File.objects.get(kf_id=kf_id)
     version = file.versions.latest("created_at")
 
-    assert set(file.valid_types) == set(
-        ["OTH", "SHM", "CLN", "DBG", "FAM"]
```
znatty22, line 26: I think the old file types also have to be removed from the file factory

znatty22 (changes requested):
Could you also test this with the branch in this PR: https://github.com/kids-first/kf-ui-data-tracker/pull/1030?

znatty22:
Thanks for taking care of this!

creator/files/migrations/0024_set_old_types_to_oth.py
gsantia: I read that using postgres you should isolate the migration using `RunPython`:

https://docs.djangoproject.com/en/3.2/ref/migration-operations/#django.db.migrations.operations.RunPython

```
Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.
```

znatty22: Oh I see

creator/files/migrations/0024_set_old_types_to_oth.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('files', '0023_add_document_types'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards_func)
```
znatty22, line 22: I think this might need the `reverse_code` populated with a no-op in case we need to roll this back for any reason:
https://docs.djangoproject.com/en/3.2/ref/migration-operations/#django.db.migrations.operations.RunPython.noop

gsantia:
> Could you also test this with the branch in this PR: [kids-first/kf-ui-data-tracker#1030](https://github.com/kids-first/kf-ui-data-tracker/pull/1030)?

What should I do?

gsantia:
Tested that the migrations run without error and change all `CLN` and `SHM` file_types to `OTH`. On the front-end, uploaded a file and set it as `Other` without error.

znatty22 (commented):
I think it looks good, but now you have to rebase bc I merged into master 😄. I can approve after that

gsantia:
> I think it looks good, but now you have to rebase bc I merged into master 😄. I can approve after that

done!

dankolbman (approved):
Looks good.

In the future, I would compress migrations into one per-app per PR. This prevents a build up of migrations which increases the test times and the dev spin up times.

znatty22 (approved):
👍

znatty22:
@dankolbman Gio didn't compress the migrations bc of this: https://github.com/kids-first/kf-api-study-creator/pull/687#discussion_r661071748

PR #690, opened 2021-07-07
🐛 Fix order of organization migrations

znatty22:
Tested in deployed dev environment with the following steps:

1. Undo add organization migrations for: buckets, referral_tokens, events, studies, and organizations
2. Run new migrations
3. Check that the Default Organization was created, and existing buckets, referral_tokens and studies are linked to the Default Organization


dankolbman (approved):
Looks like it should work. Perhaps could add the study migration dependency in addition to the initial organization migration, although the study migration already includes it.

PR #692, opened 2021-07-08
🔧 Increase max number of relay connections

### LINKED ISSUE #691: Increase connection limit
The latest bump of graphene introduced a universal enforcement of the max connection limit on edges where there was none before. This causes things that rely on many connections such as permissions in groups to only return the first 100 by default. This is ultimately an issue that should be dealt with by the client, but it will be much faster to temporarily patch by increasing `RELAY_CONNECTION_MAX_LIMIT`.

dankolbman:
Bumps the connection limit to 250 to allow single queries to return more relations.

Closes #691


gsantia (approved):
꒰⍨꒱

PR #699, opened 2021-07-16
✨ Template Field Definition Validation

### LINKED ISSUE #677: Download templates
Develop and test endpoints that will allow users to download a study's templates, the latest version of a single template, or a specific version of a template. 

Implement the endpoints in https://github.com/kids-first/kf-api-study-creator/blob/templates-rfc/docs/rfcs/0002_templates_mvp.md#-downloading-templates.

The format of the download files should be explored. It would be ideal if template field definitions and template submission files could be downloaded as one or two excel files with multiple tabs.

### LINKED ISSUE #678: Upload template field definitions
Allow users to define template field definitions in a tabular file (e.g. tsv, csv, xls, xlsx) and upload the file to update a template version's field definitions (`TemplateVersion.field_definitions`)

This could either be done by modifying the create/update template mutations in https://github.com/kids-first/kf-api-study-creator/blob/templates-rfc/docs/rfcs/0002_templates_mvp.md or creating a new upload mutation. Not sure which is best/feasible yet.

### LINKED ISSUE #679: Template field definitions
Fill in placeholders/missing functionality for management of a template's field definitions. Up until this point, `TemplateVersion.field_definitions` won't have any functionality related to it (it always defaults to an empty JSON object).

This issue should implement the specifications for template field definitions described here: https://github.com/kids-first/kf-api-study-creator/blob/templates-rfc/docs/rfcs/0003_template_fields_mvp.md 

This would include:

- Modify create template version mutation
    - Only create new template version if the field definitions have changed
- Mutation for uploading a tabular file with template field definitions 
- Validation of `TemplateVersion.field_definitions` during create/update/upload mutations
- Endpoint to serve the field definitions jsonschema??

znatty22:
Closes #679 
  
Adds validation for template field definitions via a marshmallow schema

### DIFF
--- a/creator/data_templates/field_definitions_schema.py
+++ b/creator/data_templates/field_definitions_schema.py
@@ -1,0 +1,263 @@
+from typing import Union
+
+import pandas
+from marshmallow import (
+    fields, validate, pre_load, post_load, Schema, ValidationError, EXCLUDE
+)
+from kf_lib_data_ingest.common.misc import multisplit
+from kf_lib_data_ingest.common.io import read_df
+from pprint import pprint, pformat
+
+
+def coerce_number(
+    input_value: Union[str, int, float, bool, None]
+) -> Union[int, float, str, None]:
+    """
+    Try to coerce a str literal to an int/float
+
+    If coercion is not possible return the original input value
+    """
+    # Don't try to convert None or values that are already numeric
+    # in order to preserve precision
+    if input_value is None:
+        return input_value
+
+    if isinstance(input_value, (int, float)):
+        return input_value
+
+    i_value = None
+    f_value = None
+    try:
+        f_value = float(input_value)
+        i_value = int(input_value)
+    except ValueError:
+        pass
+
+    # Input was a non-coercable str
+    if (i_value is None) and (f_value is None):
+        new_value = input_value
+    # Input was an int
+    elif i_value == f_value:
+        new_value = i_value
+    # Input was a float
+    else:
+        new_value = f_value
+
+    return new_value
+
+
+def coerce_bool(
+    input_value: Union[bool, str, int, float, None]
+) -> Union[bool, str, None]:
+    """
+    Try to coerce a str literal to a boolean
+    Coercion will include attempting to translate str synonyms to booleans
+
+    If coercion is not possible return the original input value
+    """
+    input_value = coerce_number(input_value)
+
+    if isinstance(input_value, str):
+        alternates = {
+            "true": True,
+            "false": False,
+            "yes": True,
+            "no": False,
+            "required": True,
+            "not required": False,
+        }
+        new_value = alternates.get(input_value.lower(), input_value)
+    else:
+        try:
+            new_value = bool(input_value)
+        except ValueError as e:
+            new_value = input_value
+
+    return new_value
+
+
+class FieldDefinitionSchema(Schema):
+    """
+    Marshmallow schema for a cleaning and validating a JSON object
+    containing a field definition instance in TemplateVersion.field_definitions
+
+    Notes:
+        The missing kwarg is used when the marshmallow field key is not present
+        in the input data and marshmallow needs to know if it should insert
+        the key with some kind of value
+
+        The allow_none kwarg is used when you want to allow the field to take
+        on a None value
+    """
+    key = fields.String(
+        description="Canonical name or identifier for the field",
+        missing=None,
+        allow_none=None
+    )
+    label = fields.String(
+        description="Human friendly name for the field",
+        required=True
+    )
+    data_type = fields.String(
+        description="The data type of this field's values",
+        validate=validate.OneOf(
+            ['enum', 'string', 'number', 'boolean', 'date']
+        ),
+        missing="string"
+    )
+    required = fields.Boolean(
+        description="Whether the field is required to be populated",
+        missing=False
+    )
+    accepted_values = fields.List(
+        fields.Raw(),
+        description="The list of acceptable values this field may be"
+                    "populated with",
+        # This says, if accepted_values is not None, then make sure it has
+        # at least one member in the list
+        validate=validate.Length(min=1),
+        missing=None,
+        allow_none=True
+    )
+    missing_values = fields.List(
+        fields.String(allow_none=True),
+        description="Values that a submitter may use if the data for this "
+        "field is unknown or unavailable for whatever reason",
+        # This says, if missing_values is not None, then make sure it has
+        # at least one member in the list
+        validate=validate.Length(min=1),
+
+        # If this is None, then we will interpret this to mean there are no
+        # defined missing values
+        missing=None,
+        allow_none=True
+    )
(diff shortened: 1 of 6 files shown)


creator/data_templates/field_definitions_schema.py
```diff
@@ -0,0 +1,243 @@
...
+        description="The list of acceptable values this field may be"
+                    "populated with",
+        # This says, if accepted_values is not None, then make sure it has
+        # at least one member in the list
+        validate=validate.Length(min=1),
+        missing=None,
+        allow_none=True
+    )
+    missing_values = fields.List(
+        fields.String(allow_none=True),
```
gsantia, line 115: why `fields.String` and not `fields.Raw`?

tests/data_templates/test_field_definitions.py
```diff
@@ -0,0 +1,184 @@
...
+    assert len(out_fields) == len(in_fields["fields"])
+
+    # Check that defaults were set right and all components of a field
+    # definition are present in each field definition instance
+    for out in out_fields:
+        assert set(FieldDefinitionsSchema.key_order) == set(out.keys())
+        assert out["data_type"] == "string"
+        assert out["required"] == False  # noqa
+        assert out["accepted_values"] is None
+        assert out["instructions"]is None
```
gsantia, line 184:
```suggestion
        assert out["instructions"] is None
```

gsantia (approved):

creator/data_templates/field_definitions_schema.py
```diff
@@ -0,0 +1,243 @@
...
+            else:
+                out_data[k] = v
+
+        if out_data.get("accepted_values"):
+            out_data["data_type"] = "enum"
+
+        return out_data
+
+
+class FieldDefinitionsSchema(Schema):
```
dankolbman, line 186: I would suggest adding a version number for the schema. If you do that, you can make changes to the schema in the future and then bump the version number. Thus you will have an easy way to fork logic or to identify schemas that need to be updated.

creator/data_templates/field_definitions_schema.py
znatty22: I think we always want missing values to be represented by strings (e.g. Not Available, Unknown). I can't think of why we would want to accept any other data type. For example, if i had an integer field like Age at Diagnosis and I allowed integers for missing values then how do I distinguish between real integer values vs the missing value which is also some sort of integer. For this reason, I think its best to accept only strings or null which is represented by None

creator/data_templates/field_definitions_schema.py
znatty22: I'm a little confused. I think you're suggesting adding a version to track changes to a template's field definitions? If so, do you mean add some sort of semantic version number field to the TemplateVersion model? 

If it is the latter, I'm a little hesitant to add that now bc we have no idea about who or what and how that version would be updated. Should the template admin get to update it however they see fit? Or should study creator have some rules for how to update the version (e.g. add/remove col -> major version update, desc change -> minor update)

tests/data_templates/test_field_definitions.py
gsantia: this is interesting, i guess `]` doesn't require a space after.

creator/data_templates/field_definitions_schema.py
gsantia: thanks for explaining!

tests/data_templates/test_field_definitions.py
```diff
@@ -0,0 +1,184 @@
...
+    "in_value, expected_out",
+    [
+        ("0.0", 0.0),
+        ("0", 0),
+        ("foo", "foo"),
+        ("1.234", 1.234),
+        ("200", 200),
+        (1.234, 1.234),
+        (200, 200),
+        (10.0, 10),
```
shanewilson, line 28: just curious but why does `0.0` -> `0.0` but `10.0` -> `10`?

creator/data_templates/models.py
```diff
@@ -174,8 +186,10 @@ def released(self):
...
-        Validate this template version
-
-        TODO Validate field definitions
+        Validate template version
         """
-        pass
+        # Try to fix any human errors in the data before validating against
+        # the field definitions schema (See FieldDefinitionsSchema.clean)
+        self.field_definitions = (
+            self.field_definitions_schema.load(self.field_definitions)
```
shanewilson, line 194: this happens on upload right? so we never save the file they actually give us just the new cleaned version?
Not for now just wondering for later - is it possible/easy to make a report of the values that were changed?

tests/data_templates/test_field_definitions.py
znatty22: Ha, good question. Not sure, maybe I have a bug in the code

creator/data_templates/models.py
znatty22: > so we never save the file they actually give us just the new cleaned version

Correct

> is it possible/easy to make a report of the values that were changed

yea i was thinking about this too. we could determine the values that changed in this mutation and report it back to them in the response. but perhaps more important in the near term..we may want to determine if any fields changed at all in the create new template version mutation bc we likely don't want to create a new template version unless the fields changed in some way

tests/data_templates/test_field_definitions.py
znatty22: Ok, I fixed it. It was correctly coercing "0.0" -> 0.0 and "10.0" -> 10.0 but it would convert 0.0 -> 0 and 10.0 ->10 which is probably undesired. Now it will leave ints/floats as is and only try to coerce str values. So "0.0" -> 0.0 and "10.0" ->10.0

creator/data_templates/field_definitions_schema.py
dankolbman: The version number would be authored by us whenever the schema here changes. Perhaps you add a new field to the `FieldDefinitionSchema`, then you would bump the version number. That way consumers of the data would know if they need to process the data differently.

creator/data_templates/field_definitions_schema.py
```diff
@@ -0,0 +1,251 @@
...
+    Try to coerce a str literal to an int/float
+
+    If coercion is not possible return the original input value
+    """
+    # Don't try to convert None or values that are already numeric
+    # in order to preserve precision
+    if input_value is None:
+        return input_value
+
+    if isinstance(input_value, int) or isinstance(input_value, float):
```
dankolbman, line 25: You can condense this conjunction by giving `isinstance` a tuple of types:
```python
isinstance(input_value, (int, float))
```

creator/data_templates/field_definitions_schema.py
znatty22: Ok I see. Does that mean we need a way to expose `FieldDefinitionSchema` in some way to users on the frontend? I think there is a third party library that can turn marshmallow schemas into json-schema

dankolbman (approved):
# 🥠

PR #701, opened 2021-07-20
♻️ Relax required columns in file types

### LINKED ISSUE #702: Reduce required columns in file types
Each expedited file type currently lists all of its columns as required. This is too restrictive and unnecessary as not all of the information in the required columns is actually needed. We did it this way initially, bc it was a first pass at the new KF templates and was easy to implement. 

We should reduce to the required columns for each expedited type to a subset that covers what is required by Data Service and includes fields that may not be required by Data Service but are fields we want to capture anyway. 

Example: `Clinical Sex` is a required column in the `Participant Details` expedited type. It is mapped to Data Service `participant.gender` and `participant.gender` is not required by Data Service. However, we require it in the `Participant Details` template since we would like to capture it from users.

znatty22:
Right now an expedited type includes all columns as required. Reduce the required columns for each expedited file type to a subset of columns that covers what is required by Data Service and also includes fields that may not be required by Data Service but are fields we want to capture anyway (e.g. Proband).

Closes #702

### DIFF
--- a/creator/analyses/file_types.py
+++ b/creator/analyses/file_types.py
@@ -15,89 +15,56 @@ FILE_TYPES = {
         "required_columns": [
             "First Participant ID",
             "Second Participant ID",
-            "Relationship from First to Second"
+            "Relationship from First to Second",
         ],
-        "template": "complex_family_config.py"
+        "template": "complex_family_config.py",
     },
     "FTR": {
         "name": "Family Trio",
         "required_columns": [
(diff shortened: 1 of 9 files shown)


chris-s-friedman (approved):
Thanks @znatty22 ! looks good!

dankolbman (approved):
# 🚜

gsantia (approved):
🍉

PR #703, opened 2021-07-20
✨ Add User organization filter

### LINKED ISSUE #700: Filter allUsers by organization
The `allUsers` query should accept an option to filter users for a given organization.

gsantia:
Closes #700 

Add the ability to filter users by organization in the `allUsers` query. An organization node ID is supplied as input to the query and all users that are a member of that organization are returned.

Screen Shot 2021-07-23 at 4 21 04 PM

### DIFF
--- a/creator/organizations/mutations.py
+++ b/creator/organizations/mutations.py
@@ -5,7 +5,7 @@ from graphql_relay import from_global_id
 from django.contrib.auth import get_user_model
 
 from creator.organizations.models import Organization
-from creator.organizations.nodes import OrganizationNode
+from creator.organizations.queries import OrganizationNode
 from creator.events.models import Event
 
 User = get_user_model()
--- a/creator/users/schema.py
+++ b/creator/users/schema.py
@@ -1,25 +1,18 @@
 import graphene
-import django_filters
 from graphene import relay
 from graphql_relay import from_global_id
 from graphql import GraphQLError
 from graphene_django import DjangoObjectType
 from graphene_django.converter import convert_django_field_with_choices
-from graphene_django.filter import (
-    DjangoFilterConnectionField,
-    GlobalIDFilter,
-)
-from django_filters import OrderingFilter
+from graphene_django.filter import DjangoFilterConnectionField
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
 
 from creator.studies.models import Membership
-from creator.users.mutations import Mutation
-
-from creator.organizations.nodes import OrganizationNode
 from creator.organizations.queries import OrganizationFilter
-
+from creator.users.mutations import Mutation
+from creator.users.queries import Query
 
 User = get_user_model()
 
@@ -100,7 +93,7 @@ class UserNode(DjangoObjectType):
     display_name = graphene.String(source="display_name")
 
     organizations = DjangoFilterConnectionField(
-        OrganizationNode,
+        "creator.organizations.queries.OrganizationNode",
         filterset_class=OrganizationFilter,
         description="List all organizations the user is in",
     )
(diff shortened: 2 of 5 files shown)


tests/users/test_queries.py
```diff
@@ -0,0 +1,115 @@
...
+from django.contrib.auth import get_user_model
+from creator.organizations.factories import OrganizationFactory
+from creator.organizations.models import Organization
+from creator.users.factories import UserFactory
+
+User = get_user_model()
+
+ALL_USERS_ORGANIZATION = """
+query ($org_name: String) {
+  allUsers (organizations_Name: $org_name) {
```
shanewilson, line 13: the allStudies query uses an id for organization filters - if might be good to do that here as well:

```
allStudies(organization: ID!) 
```

creator/users/schema.py
```diff
@@ -93,6 +95,12 @@ def _get_study_id(self, info):
 class UserNode(DjangoObjectType):
     display_name = graphene.String(source="display_name")
 
+    organizations = DjangoFilterConnectionField(
```
shanewilson, line 98: this is good - can we also add `users` sub query to an `organization`?

tests/users/test_queries.py
gsantia: yeah, i wasn't sure about this

creator/users/schema.py
gsantia: sure, i'll make an issue and work on it and try to get it out today

creator/users/schema.py
```diff
@@ -189,6 +197,7 @@ class Meta:
             "last_name": ["exact"],
             "last_login": ["gt", "lt"],
             "date_joined": ["gt", "lt"],
+            "organizations__name": ["exact"],
```
dankolbman, line 200: As you can see, graphene converts these to strangely formatted parameters like `organizations_Name`. I'd like to avoid these by explicitly defining the fields on the filter where you can control the name.

tests/users/test_queries.py
dankolbman: This should be supported by simply plugging in `organization` into the filterset. If not, you can use `GlobalIDMultipleChoiceFilter` to create the filter field.

dankolbman (approved):
# 🛺

tests/users/test_queries.py
```diff
@@ -0,0 +1,117 @@
...
+
+from django.contrib.auth import get_user_model
+from creator.organizations.factories import OrganizationFactory
+from creator.organizations.models import Organization
+from creator.users.factories import UserFactory
+
+User = get_user_model()
+
+ALL_USERS_ORGANIZATION = """
+query ($organizations: [ID]) {
```
shanewilson, line 13: can we make this type `$organizations: [ID!]`? If you add a value to the list we shouldn't allowed it to be `null`

tests/users/test_queries.py
```diff
@@ -0,0 +1,117 @@
...
+  allUsers (organizations: $organizations){
```
shanewilson, line 14: just a note - allStudies only allows a single organization id. Allowing a list is probably more flexible, but we should make sure to align the other endpoints to match.

gsantia: honestly, i wasn't able to figure out how to do it in the case of a single org id, because it's a many-to-many relationship. i agree, i can add that functionality to `allStudies`

dankolbman: I'd agree with being consistent with other queries. Allowing to filter by multiple orgs would be nice, but we don't have that need right now and it's not how we do things elsewhere. You can probably achieve this pretty easily by swapping out to `GlobalIDFilter` instead of the multiple choice.

creator/users/schema.py
```diff
@@ -180,6 +192,8 @@ def get_node(cls, info, id):
 
 
 class UserFilter(django_filters.FilterSet):
+    organizations = GlobalIDFilter(field_name="organizations")
```
shanewilson, line 195: i think these should be `organization` now right?

```suggestion
    organization = GlobalIDFilter(field_name="organization")
```

gsantia: true

shanewilson (approved):

tests/users/test_queries.py
```diff
@@ -0,0 +1,117 @@
...
+def test_query_all_users_organization(
+    db, permission_client, permissions, allowed
+):
+    """
+    Test the allUsers query for filtering by organization name functionality.
+    """
+    user, client = permission_client(permissions)
+
+    # Setup users and organizations
+    org1 = OrganizationFactory()
```
znatty22, line 49:
```suggestion
    org1, org2, org3 = OrganizationFactory.create_batch(3)
```

tests/users/test_queries.py
```diff
@@ -0,0 +1,117 @@
...
+    org2 = OrganizationFactory()
+    org3 = OrganizationFactory()
+
+    user1 = UserFactory()
```
znatty22, line 53:
```suggestion
    user1, user2, user3 = UserFactory.create_batch(3)
```

tests/users/test_queries.py
```diff
@@ -0,0 +1,117 @@
...
+    user2 = UserFactory()
+    user3 = UserFactory()
+    for _ in range(10):
```
znatty22, line 56:
```suggestion
    UserFactory.create_batch(10)
```

znatty22 (changes requested):

gsantia:
Update on this, I've been stuck with circular imports lately and cannot figure out how to solve them without refactoring organizations and users. In another PR I added a `users` feature to `organization` queries, letting you see all the users in an organization. This required adding `UserNode` and `UserFilter` to the `OrganizationNode` definition.

But then in this PR we want to basically do the opposite, and we need `OrganizationNode` and `OrganizationFilter` to show up in the definition of `UserNode`. This leads to circular imports. At this point I've tried every trick I can think of and also whatever suggestions I can find online. Delayed imports don't work because they're both class definitions which execute immediately. Am I missing something? Otherwise I can do a bit of refactoring here and present it.

Basically I'd need to define `OrganizationNode` and `UserNode` in the same file and make all the changes that would entail. Not sure where this file would live. Probably in the organization app? Does this make sense?

shanewilson:
> Update on this, I've been stuck with circular imports lately and cannot figure out how to solve them without refactoring organizations and users. In another PR I added a `users` feature to `organization` queries, letting you see all the users in an organization. This required adding `UserNode` and `UserFilter` to the `OrganizationNode` definition.
> 
> But then in this PR we want to basically do the opposite, and we need `OrganizationNode` and `OrganizationFilter` to show up in the definition of `UserNode`. This leads to circular imports. At this point I've tried every trick I can think of and also whatever suggestions I can find online. Delayed imports don't work because they're both class definitions which execute immediately. The Django importing with strings pattern also doesn't work because these are not Model classes. Am I missing something? Otherwise I can do a bit of refactoring here and present it.
> 
> Basically I'd need to define `OrganizationNode` and `UserNode` in the same file and make all the changes that would entail. Not sure where this file would live. Probably in the organization app? Does this make sense?

hey @gsantia you should be able to solve this by either using a "string" or a lambda. For example here https://github.com/kids-first/kf-api-study-creator/pull/703/files#diff-c4b79ed7497d7b73defcc01c8ca2a43aed837b3cd4a5c678e923ceffbdc00401R103 if you change it to the string `"OrganizationNode"` graphene should lazy load the model. If that doesn't work you may need to use the full path like `"creator.organizations.nodes.OrganizationNode"`.

https://stackoverflow.com/questions/62773292/graphql-design-with-circular-dependency
https://github.com/graphql-python/graphene/blob/a53b782bf8ec5612d5cceb582fbde68eeba859aa/graphene/utils/module_loading.py#L5

gsantia:
> hey @gsantia you should be able to solve this by either using a "string" or a lambda. For example here https://github.com/kids-first/kf-api-study-creator/pull/703/files#diff-c4b79ed7497d7b73defcc01c8ca2a43aed837b3cd4a5c678e923ceffbdc00401R103 if you change it to the string `"OrganizationNode"` graphene should lazy load the model. If that doesn't work you may need to use the full path like `"creator.organizations.nodes.OrganizationNode"`.
> 
> https://stackoverflow.com/questions/62773292/graphql-design-with-circular-dependency
> https://github.com/graphql-python/graphene/blob/a53b782bf8ec5612d5cceb582fbde68eeba859aa/graphene/utils/module_loading.py#L5
> 
> if none of that works try `lambda: OrganizationNode` or `lambda: creator.organizations.nodes.OrganizationNode`

thanks for the suggestions! unfortunately i've tried these ideas and several variations of them and it still won't work. i think the key issue is that organization needs to use `UserFilter` now and user needs to use `OrganizationFilter`. so they're more tightly-coupled than just between node definitions. i'm trying to put them all into one file right now.

shanewilson:
Are you able to use a "string" or a lambda on the Filters as well to decouple them? If we can't this seems like a major regression in Graphene since I know we used to be able to handle this in an earlier version.

gsantia:
> Are you able to use a "string" or a lambda on the Filters as well to decouple them? If we can't this seems like a major regression in Graphene since I know we used to be able to handle this in an earlier version.

maybe I'm doing it wrong, but it doesn't seem to be working. i think i found a solution though and will push it soon

creator/organizations/mutations.py
```diff
@@ -5,14 +5,14 @@
...
-from creator.organizations.nodes import OrganizationNode
+from creator.organizations.queries import OrganizationNode
 from creator.events.models import Event
 
 User = get_user_model()
 
 
 class CreateOrganizationInput(graphene.InputObjectType):
-    """ Schema for creating a new study """
+    """Schema for creating a new study"""
```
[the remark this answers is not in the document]
gsantia: yeah sorry it was black

dankolbman (approved):
# 🍙

znatty22 (approved):
👍

PR #704, opened 2021-07-21
📌 Pin ecs_service_type_1 version to avoid tagging issue

znatty22:
Bypass Jenkins library change for ecs tag format issue. More details here: https://github.com/kids-first/aws-ecs-service-type-1-module/pull/249


dankolbman (approved):
# 🛻

znatty22:
Looks like we don't need this. @blackdenc made a patch that seems to resolve the underlying issue

PR #711, opened 2021-07-26
✨ Upload field definitions

### LINKED ISSUE #678: Upload template field definitions

znatty22:
Closes #678

Upload a template's field definitions in a tabular or json file
Also includes a `faketemplates` command that adds some realistic templates to all studies within the provided org

image


gsantia:
got some version event tests failing

znatty22:
Huh..i have no idea why those tests are failing..i didn't do anything with events

tests/data_templates/test_upload_fields.py
```diff
@@ -0,0 +1,144 @@
...
+        json_file.seek(0)
+        data = {
+            "operations": json.dumps(
+                {
+                    "query": UPLOAD_FIELDS_MUTATION.strip(),
+                    "variables":  {"file": None}
+                }
+            ),
+            "file": json_file,
+            "map": json.dumps({"file": ["variables.file"]}),
```
gsantia, line 64: what is this?

creator/data_templates/management/commands/faketemplates.py
```diff
@@ -0,0 +1,109 @@
...
+        org = Organization.objects.filter(name=org_name).first()
+        if not org:
+            raise Organization.DoesNotExist(
+                f"Organization with name {org_name} does not exist!"
+            )
+
+        if options["delete"]:
+            count, _ = org.data_templates.all().delete()
+            logger.info(
+                f"Deleted current templates in organization {count}"
```
gsantia, line 110:
```suggestion
                f"Deleted {count} current templates in organization {org.name}"
```

tests/data_templates/test_upload_fields.py
znatty22: Ugh yea its weird. Its part of the graphql multipart request spec: https://github.com/jaydenseric/graphql-multipart-request-spec#single-file. 

I think the `map` value is basically saying look at the `file` variable in the `variables` part of the request content in order to read the file content from the file stream

gsantia (approved):

creator/data_templates/mutations/template_version.py
```diff
@@ -277,6 +285,64 @@ def mutate(self, info, id):
...
+            raise GraphQLError(
+                f"File size {file_size} exceeds max size: "
+                f"{settings.FILE_MAX_SIZE}"
+            )
+
+        try:
+            field_definitions = TemplateVersion.load_field_definitions(
+                file, original_name=file.name
+            )
+        except ValidationError:
```
dankolbman, line 330: Why do you need to specifically filter out `ValidationError` instead of letting it be handled by the catch all below?

znatty22: The catch-all replaces the original exception message with a general one so the user wouldn't get the details of the validation error. I don't think I want the user to see details of every exception so that's why I logged the exception in the catch-all so the stack trace and exception message appear in the app logs, while the the GraphQLError just returns a simple something-went-wrong message to the user on the frontend

dankolbman (approved):
# 🥘

gsantia (approved):

PR #713, opened 2021-07-28
✨ Validate file version against template

### LINKED ISSUE #694: Refactor File Upload Validation
Once the templates have been implemented, a user will be able to upload a file version and select which template or file type they believe their file to be. Once they submit their selection, study creator will analyze the file content and evaluate how complete the file is according to it's template. 

A file version with a selected file type besides OTH, should not complete the creation process unless it has all of the elements required by its file type.

Visual details can be found in the [Data Template User Flows](https://docs.google.com/presentation/d/101n1uKvORAuXjCG8BiBYDFgfwkvpW5NqXqsXdn4s6xY/edit#slide=id.ge3f7456eee_0_275)

image

znatty22:
Closes #694 

Introduces a new mutation that will validate a file version against a user selected template. This will be used to refactor the file submission flow in Data Tracker according to https://github.com/kids-first/kf-ui-data-tracker/issues/1029.


creator/files/mutations/version.py
```diff
@@ -219,8 +223,132 @@ def mutate(
...
+    )
+    study = graphene.ID(
+        description="The study which the file version belongs to. Will be "
+        "used to find the study templates that match the file version. "
+        "Either the template_version or study must be provided to "
+        "validate the file version."
+    )
+
+
+class ValidateFileVersionMutation(graphene.Mutation):
```
dankolbman, line 245: I might recommend splitting this into two mutations, maybe `ValidateTemplateMatch` and `SuggestTemplates`.  This way, the one mutation isn't overloaded into multiple forks that split on the provided parameters (not immediately evident to the client what's happening) and the client doesn't need to determine what happened based on the shape of the results (this is acting against the typing that graphql provides rather than with it).

If you have a `ValidateTemplateMatch`, you can very clearly tell the client to expect a True/False value for success.
If you have a `SuggestTemplates`, you can very clearly tell the client to expect a list of templates, perhaps ordered or with additional information.

Combining the two means you have to fork on the input parameters and the client will have to fork on the result whereas it would be better if both you and the client could just do one thing didn't have to make logic forks.

creator/files/mutations/version.py
```diff
@@ -219,8 +224,82 @@ def mutate(
...
+    """
+    class Arguments:
+        input = EvaluateTemplateMatchInput(
+            required=True,
+            description="Attributes for the mutation",
+        )
+
+    results = graphene.List(TemplateMatchResult)
+
+    def mutate(self, info, input):
```
znatty22, line 257: Just realized I should check if user has permission to view file and study in this mutation. Will implement once I return to my comp

creator/files/models.py
```diff
@@ -92,13 +93,19 @@ class Meta:
...
         help_text="The user who originally created this File",
     )
-
+    template_version = models.ForeignKey(
+        TemplateVersion,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="files",
+        help_text="The data template version this file conforms to",
```
gsantia, line 103: small thing, when this field is populated will we always know that the file conforms to the template_version?
otherwise, could say `The data template version this file should conform to`

znatty22: Actually, I'm not going to change it bc it would generate a new migration. Then I'd have to go through the annoying process of logging into my Dev EC2 and undoing the current migration and then redeploying

creator/files/utils.py
```diff
@@ -0,0 +1,89 @@
...
+def evaluate_template_match(file_version, template_version):
+    """
+    Evaluate whether a file version matches a template and report the details
+    of the mismatch between the file's columns and the template's
+    required and optional columns.
+
+    The file "matches" a template if the file contains all of the required
+    columns in the template. If the template does not have any required
+    columns defined, then a file "matches" a template if it contains all
+    of the optional columns in the template.
```
gsantia, line 38: what's the purpose of a template with no required columns?

creator/files/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    results["matched_optional_cols"] = optional_cols & file_columns
+    results["missing_optional_cols"] = optional_cols - file_columns
+    for k, v in results.items():
+        results[k] = list(v)
+
+    # If the template has required cols, then use this to determine match
+    if required_cols:
+        results["matches_template"] = (
+            len(results["missing_required_cols"]) == 0
+        )
```
gsantia, line 82: totally minor,

```suggestion
results["matches_template"] = (
            bool(results["missing_required_cols"])
        )
```

PR #714, opened 2021-08-02
Replace fakestudies and fakefiles with a fakeorgs command

### LINKED ISSUE #649: Replace fakestudies and fakefiles with a fakeorgs command
In order to bootstrap mock data, we currently run `fakestudies` followed by `fakefiles`. This creates a handful of fake studies and then will add a bunch of fake files to those studies.
This is a little non-sensible because the study factory should call on the file factory to create studies whenever a fake study is created. Furthermore, now with the idea of organizations, the organization factory should create studies which will create files. If these factories are setup correctly, we should be able to create a simple `fakeorgs` command which will call `OrganizationFactory.create_batch(n)` which will take care of all the mock data setup.

gsantia:
Closes #649


gsantia:
I am having issues getting the `OrganizationFactory` to work. The problem stems from the fact that when populating the `organization` field for `Study`, the `StudyFactory` calls `OrganizationFactory`.

When an `Organization` is created, a number of `Studies` are automatically created, which in turn call `OrganizationFactory` again. Usually this is solved by using the `@factory.post_generation` decorator to perform the desired operations after an object is created, or by using `organization=factory.RelatedFactory()` for a similar result. This doesn't work here because `organization` is a required field for `Study`, and we get errors like:
```
django.db.utils.IntegrityError: null value in column "organization_id" violates not-null constraint
DETAIL:  Failing row contains (SD_VYYOW7JL, transform extensible e-commerce, t, 2021-08-11 16:19:45.62252+00, he-push-exactly, null, null, walk-them-program, 2021-08-11 16:19:45.622626+00, null, null, null, null, , null, , f, UNKNOWN, UNKNOWN, UNKNOWN, null, t, null, {}, null).
```

Doing `organization=factory.SubFactory()` will populate the `organization` field as required, but then we just get infinite recursion:
```
RecursionError: maximum recursion depth exceeded
```

Is there some way to manually overload the `organization` attribute when we want to? Based on what I've read, I felt this should work:
```
@factory.post_generation
    def studies(self, create, extracted, **kwargs):
        """
        After an organization is created with the factory, generate _extracted_
        related Studies. If _extracted_ is not provided, default to
        _DEFAULT_NUM_.
        """
        if not create:
            return
        DEFAULT_NUM = 2
        extracted = DEFAULT_NUM if extracted is None else extracted
        if extracted:
            StudyFactory.create_batch(extracted, organization=self)
```
for creation of the `Studies` of an `Organization`, but it doesn't.

znatty22:
Closing due to inactivity

PR #715, opened 2021-08-04
🛠 add 'sentry-trace' to CORS allowed headers

jgnieuwhof:
## Motivation
We're using Sentry in the new DeWrangle app.

Unfortunately we've been encountering some CORS related errors:
- https://github.com/kids-first/dewrangle/pull/26/files#diff-882b2c04b01b2e8b2cdcf1817c30ea503a8005f1c0d54cfff37053b6801fea85R29-R33

## Approach
Sentry uses the `sentry-trace` header to correlate issues across the front and back end pieces of an application:
- https://develop.sentry.dev/sdk/performance/#header-sentry-trace

Since this is a non-standard header we need to configure the CORS middleware to allow it:
- https://pypi.org/project/django-cors-headers/

I've made this change for both local development, and production, since Sentry is expected to work in both.

### DIFF
--- a/creator/settings/development.py
+++ b/creator/settings/development.py
@@ -37,7 +37,7 @@ STAGE = "dev"
 ALLOWED_HOSTS = ["*"]
 CORS_ORIGIN_ALLOW_ALL = True
 CORS_ALLOW_HEADERS = list(default_headers) + [
-    'sentry-trace',
+    "sentry-trace",
 ]
 
 DEVELOP = True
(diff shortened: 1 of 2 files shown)


shanewilson (commented):
Looks good! but I'll let the python devs give feedback

dankolbman (changes requested):
Looks good! Can you just make sure to run it through the black formatter first? (line length 80)

jgnieuwhof:
@dankolbman - sure thing; formatted both files via the command `black -l 80 <filename>`.

dankolbman (changes requested):
# 🌵

Looks good!

dankolbman (approved):

gsantia (approved):
🔮

PR #716, opened 2021-08-05
📝 Add RFC for storage backends

dankolbman:
Add an RFC for details surrounding Storage Backends.

### DIFF
--- a/docs/rfcs/XXXX_storage_backends.md
+++ b/docs/rfcs/XXXX_storage_backends.md
@@ -1,0 +1,288 @@
+# Storage Backends
+
+**Status:** Approved, waiting for implementation
+
+
+## Background and Existing Solution
+
+![enter image description here](https://i.ibb.co/C7bzPNZ/storage-layout.png)
+> A simple representation of how the application and its resources are currently allocated.
+
+The the application relies on a data storage system to both save files that it receives and to account for other files that have been deposited externally. The current method that is used is adequate for a single tenant operating in a single AWS account, but needs to be re-thought for multiple tenants that  have more sophisticated storage needs.
+
+### Study Buckets
+
+Studies are central entities in the the application. They partition data that is collected and submitted by different research projects and investigators. For this reason, it was determined that each study would have its own bucket to segment it from other collections of data.
+
+#### Creation
+
+The the application automatically creates and configures buckets upon creation of a new study. These buckets are created in the same account as the IAM role that the API uses (Kids First Open Data). This account happens to be different from the account where the API actually resides (Kids First STRIDES). This is necessary as compute used to be contained within Kids First Open Data but was migrated to the STRIDES account. However, the data did not move and the codebase required that the IAM used was owned by the same account as the data.
+
+#### Configuration
+
+Each study bucket needs to be configured with the correct security policies, encryption, replication, logging, and so on. For this reason, the bucket creation process was automated and included as part of the the application's service offerings. Upon creation, the the application will configure these additional settings ensuring all buckets are setup correctly.
+
+#### Layout
+Each study bucket has a similar layout:
+```
+source/
+  uploads/
+harmonized/
+```
+
+Under `source` are data files such as raw read files received from the sequencing center, MRI DICOMs from the investigator, and so forth. The `uploads` directory contains files that are uploaded from the the application. `harmonized` contains files that were processed by the BIX team using things such as a standard alignment workflow or calling pipeline.
+
+### Issues with Study Buckets
+
+Study buckets are restricted to the same account as the IAM profile assigned to the application's compute instance.
+The IAM profile contains a large surface of permissions needed to create and modify buckets.
+
+## Storage Backend Solution
+By allowing users to create "Storage Backends" that grant the application access to bucket resources in other accounts, we will be able to provide more flexibility in how organizations choose to store their data.
+
+## Default Storage
+![enter image description here](https://i.ibb.co/vBb1YsN/storage-layout.png)
+"Default Storage" will refer to the defaults for how a new organization and its studies will be stored. It's important that new organizations and studies be quickly bootstraped into a working state without unnecessary configuration. To achieve this, the application will provide a default workspace for documents to land. This will be achieved by providing a Storage Backend that will point to the application's application bucket.
+
+### Application Bucket
+
+It's proposed that the application maintain a single bucket in which it may stash data. This bucket will be intended to contain anything the the application may need to store such as configuration files, user avatars, and similar. Most notably, though, it will contain uploaded documents for each study. In this way, the single application bucket will deprecate the need for each study bucket to contain its uploads. However, it is also recognized that perhaps some organizations may wish their uploaded documents to reside within their own resource pool. For this case, we will allow the document bucket to be specified within a different account.
+The default solution for a new organization (and the existing Kids First organization) will be to store all documents within the application bucket. These documents will be kept as outlined above in the `New Layout` where the `studies/` prefix will be at the root of the bucket.
+
+### New Layout
+
+The layout of files uploaded by the application is outlined below. This structure will ultimately be placed within the storage backend's bucket under the provided prefix:
+```
+dewrangle/
+	studies/
+		SD_00000000/
+			source/
+			uploads/
+			harmonized/
+```
+
+## External Storage Backends
+
+![enter image description here](https://i.ibb.co/PTPMLnf/storage-layout.png)
+When/ an organization requires all of their uploads or data files to be under their control, an organization administrator may specify credentials and details for a bucket of their choosing in a different account.
+
+### Storage Backends
+
+A Storage Backend is a simple entity that belongs to an organization and specifies IAM user credentials, a bucket that the IAM user has access to, and a prefix within that bucket. This concept will allow the application to access buckets across different accounts for the purpose of storing or evaluating files.
+
+### Configuration
+
+To configure a storage backend, an organization administrator will be required to provide details about the AWS account where their target bucket resides and a user who may access it. An organization administrator will be able to configure this by navigating to the *Organization Settings* screen and choosing *Storages*. This will allow the administrator to add a new storage by first asking the user to create a new IAM user in their AWS account and assigning a specified policy. After the user has done this, they will be required to provide the AWS access and secret key pair so that the application may access the bucket. The final step will be to allow the user to specify the bucket and what prefix under it they would like this storage backend to use as the root. After confirming their choices, the application will attempt to access the bucket with the provided credentials and save the new storage backend if successful.
+
+### Use of Storage Backends
+
+A Storage Backend may be used in one of two ways: Either as a Document Storage Backend or a Data Storage Backend. Each method will require the same configuration of a Storage Backend and, in fact, a single Storage Backend may be used for both Documents and Data.
+
+#### Document Storage Backends
+
+Storage Backends used for documents will be read/write. the application will use these backends to save documents using a `dewrangle/uploads/` prefix under the root of the Storage Backend.
+
+#### Data Storage Backends
+
+Storage Backends used for data will only utilize read and list operations. These backends will be used for file discovery and accounting only.
+
+### Setting Organization Storage Backend Defaults
+
+If an organization administrator wishes all new studies to use a given backend for document uploads or data files, they may specify it in the organization settings. The administrator will have two defaults to configure: *Default Document Storage Backend* and *Default Data Storage Backend*. An administrator may use the same Storage Backend for both, but it will be up to them to prevent collisions.
+
+### Changing Storage Backends Per-Study
+
+It may be desirable for a study administrator to use different Storage Backends for different studies. For this, studies will be able to specify one Storage Backend for documents and one or more Storage Backends for data files. 
+
+#### Kids First Example
+
+In Kids First, each study keeps its data in different buckets including documents. If we were to use the Storage Backend approach for Kids First, we could do so by first creating a new Storage Backend for each bucket and then updating each study with the appropriate Storage Backend.
+If we would like to stop distributing uploaded documents to each study's bucket, we could simply update each study's Document Storage Backend to use the Default Storage Backend. This would result in all uploaded documents going to the application's bucket with all data operations occurring on each individual study's Data Storage Backend.
+
+### Management of External Buckets
+
+the application will not be responsible for configuration of the external buckets. This responsibility resides with the organization administrator and whoever controls their resources.
+
+## User Flows
+
+### Creating a Storage Backend
+
+Go to organiztion study page and add a new Storage Backend
+
+User provides basic information such as the bucket name and what prefix they want to keep data under
+![](https://i.ibb.co/KwZbzdr/Basic.png)
+
+The user will be asked to configure an IAM user within their AWS account that the API may use to interact with the specified bucket.
+![](https://i.ibb.co/ys5xc61/Create-User.png)
+
+The user will need to create a new programmatic IAM user within the AWS dashboard
+
+![](https://i.ibb.co/7bVWVFR/Screenshot-2021-08-10-at-15-47-08-IAM-Management-Console.png)
+
+The policy that is suggested or one that allows for similar access will need to be created to be attached to the new IAM user
+![](https://i.ibb.co/zQhnYT5/Screenshot-2021-08-10-at-15-51-49-IAM-Management-Console.png)
+
+The new IAM user can then be created
+![](https://i.ibb.co/zhj02jx/Screenshot-2021-08-10-at-15-52-42-IAM-Management-Console.png)
+
+In the application, the user may now paste in the credentials for the new IAM user
+![](https://i.ibb.co/yBXbjTJ/Create-User-1.png)
+
+Once this is done, the application will confirm that it has the permissions necessary to perform its operations
+
+![](https://i.ibb.co/1fk2B48/Create-User-3.png)
+
+The new storage backend will now be available for use in studies within the organization.
+
+
+### Modifying Storage Backends for Studies
+
+Study administrators may change a study's settings to use any Storage Backend that is in the organization.
+![](https://i.ibb.co/DQZTKCr/Screen-Shot-2021-08-10-at-4-16-15-PM.png)
+
+A study may only have one Document Storage Backend to read and write uploaded documents to. If this location changes after uploads have been made, the application will not be able to find old files unless they are migrated to the correct location in the new backend and updated in the application database. It may be possible to automate this in the future.
+
+[========]
+
+
+## Technical Implementation
+
+### StorageBackend Model
+
+The storage backend model
+
+```graphql
+type StorageBackend {
+  id: ID!
+  name: String!
+  bucket: String
+  prefix: String
+  organization: Organization
+  accessKey: String
+  secretKey: String
+  arn: String
+  account_id: Int
+  account_alias: String
+  user: String
+}
+```
+
+#### Secret Storage
+
+The `access_key` and `secret_key` will be stored in the application's database in an encrypted state. The encryption will be done with a KMS key provisioned specifically for the application. In order to perform actions on behalf of the IAM user and thus the stored key pair, the secrets must be decrypted first using the same KMS key.
+
+#### Blank Secrets
+
+The secret key pair is *optional*. This is to allow other access methods such as direct modification of the application's IAM user.
+For example, the application's IAM user may already be granted access to a bucket location through its policy. To create a storage backend that utilizes this location requires only that the application be informed of the the bucket and prefix. No new access credentials will be required as the application will use its IAM user to try to access the location by default.
+
+### Mutations
+
+#### Create
+
+```graphql
+type CreateStorageBackendInput {
+  name: String!
+  bucket: String!
+  prefix: String
+  organization: ID!
+  accessKey: String
+  secretKey: String
+  healthy: Bool!
+}
+createStorageBackend(input: CreateStorageBackendInput): StorageBackend!
+```
+Only organization administrators may be allowed to create new storage backends.
+The create mutation includes an attempt to verify that the application has proper access to the specified resource and desired actions.
+
+#### Update
+
+```graphql
+type UpdateStorageBackendInput {
+  name: String!
+  accessKey: String
+  secretKey: String
+}
+updateStorageBackend(id: ID!, input: UpdateStorageBackendInput): StorageBackend!
+```
+Only organization administrators may be allowed to update Storage Backends. If the secrets are being updated, the validity of the access will be checked.
+
+#### Delete
+
+```graphql
+deleteStorageBackend(id: ID!): StorageBackend!
+```
+Only Organization administrators may be allowed to delete existing storage backends in their Organization. This will remove the Storage Backend and its secrets entirely from the database. Any studies that use the Storage Backend will be reset to the default the application Storage.
+
+#### Update Storages
+
+```graphql
+type UpdateStoragesInput {
+  documentBucket: ID
+  dataBuckets: [ID]
+}
+updateStorages(id: ID!, input: UpdateStorageBackendInput): Study!
+```
+
+This mutation allows a user to change the Storage Backend configuration for a study.
+
+### Updates to the Study Model
+
+The study model will take on two new relationships:
+
+```graphql
+type Study {
+  ...existing study fields
+  documentStorage: StorageBackend
+  dataStorage: [StorageBackend]
+}
+```
+
+### Updates to Study Creation
+
+Because all future studies will utilize the default storage backend, it will no longer be necessary to setup new buckets for each study. To disable this behavior, the `FEAT_STUDY_BUCKETS_CREATE_BUCKETS` setting may simply be set to `False` or we may choose to make it an option in the mutation or remove the code alltogether.
+
+### Updates to Download Views
+
+Download views will first need to lookup a requested file's Storage Backend through association with its study. This will then require decrypting secrets and using the IAM user to retrieve the requested file.
+
+### Updates to Upload Mutations
+
+Upload mutations for new documents and versions will need to be updated to retrieve the related Storage Backend's credentials to write files as needed.
+
+### Updates to Other File Access
+
+> TODO
+>
+> File access should hopefully be transparant through implementing a custom Django storage backend that will retrieve credentials automatically making any necessary file access work seamlessly.
+
+### Default Storage
+
+A default storage location will be provisioned for each Organization in the application. This location will reside within the application bucket provisioned for the application. A prefix will be designated for the organization and the access credentials will be blank as the application's IAM policy will allow the correct access.
+
+### Periodic Maintenance
+
+It may be desireable to have a daily occuring task that will perform verification that each Storage Backend has keys that are actively attached to a user with the desired access. If a Storage Backend ever has issues comunicating with the provided credentials, it may be most helpful to send an automated email to an Organization administrator to remedy the problem.
+
+## Migrating
+
+### Existing Study Buckets
+
+Existing study buckets will need to be added as storage backends in the application. These storage backends will not require new IAM users to be created as the application's IAM profile should still contain the permissions needed to access them. from here, existing files should be retrievable to the application in their original location although new uploads may have the new root prefix prepended (`dewrangle/uploads/`)
+
+### Moving to the Default Storage
+
+It may be desirable to relocate all documents uploaded to the application within the default storage. To do this, the following would need to be performed in order:
+
+1 - Copy files out of `uploads` directories in study buckets to `dewrangle/<study_id>/uploads/`
+2 - Update paths of existing files in database to point to new location
+3 - Remove files from old upload location in study buckets
+
+
+### Relocating a Study to a New Storage Backend
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+```graphql
+type StorageBackend {
+  id: ID!
+  name: String!
+  bucket: String
+  prefix: String
+  organization: Organization
+  accessKey: String
+  secretKey: String
+  studies: [one-many]
```
znatty22, line 155: I think this would be many-to-many? Bc a study can have both a document storage backend and data storage backend and those backends may be associated with multiple studies

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+}
+```
+
+#### Secret Storage
+
+The `access_key` and `secret_key` will be stored in the application's database in an encrypted state. The encryption will be done with a KMS key provisioned specifically for Dewrangle. In order to perform actions on behalf of the IAM user and thus the stored key pair, the secrets must be decrypted first using the same KMS key.
+
+#### Blank Secrets
+
+The secret key pair is *optional*. This is to allow other access methods such as direct modification of the application's IAM user.
```
znatty22, line 165: Can you explain this a bit more? I'm not understanding why the secret key pair could be optional

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+
+## Technical Implementation
+
+### StorageBackend Model
+
+The storage backend model
+
+```graphql
+type StorageBackend {
```
znatty22, line 147: I wonder if the IAM User should actually be its own entity with a relationship to the storage backend. Bc the same IAM user may have access to different buckets in the same account. You probably don't want to store the username, access key, secret key multiple times right? If the key pair changes then you have to update it for every storage backend

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+  id: ID!
+  name: String!
+  bucket: String
+  prefix: String
+  organization: Organization
+  accessKey: String
```
znatty22, line 153: Is this missing the IAM username? Might be useful to store for info/display purposes

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+  secretKey: String
+  studies: [one-many]
+}
+```
+
+#### Secret Storage
+
+The `access_key` and `secret_key` will be stored in the application's database in an encrypted state. The encryption will be done with a KMS key provisioned specifically for Dewrangle. In order to perform actions on behalf of the IAM user and thus the stored key pair, the secrets must be decrypted first using the same KMS key.
```
znatty22, line 161: What are the pros/cons of storing it in the db vs in the application's secrets bucket where we currently store our secret key pair?

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+```graphql
+type CreateStorageBackendInput {
+  name: String!
+  bucket: String!
+  prefix: String
+  organization: ID!
+  accessKey: String
+  secretKey: String
+  healthy: Bool!
```
znatty22, line 179: What is this boolean used for?

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Status:** Approved, waiting for implementation
+
+
+## Background and Existing Solution
+
+![enter image description here](https://i.ibb.co/C7bzPNZ/storage-layout.png)
+> A simple representation of how Data Tracker and its resources are currently allocated.
+
+The Data Tracker relies on a data storage system to both save files that it receives and to account for other files that have been deposited externally. The current method that is used is adequate for a single tenant operating in a single AWS account, but needs to be re-thought for multiple tenants that  have more sophisticated storage needs.
```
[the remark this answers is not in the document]
dankolbman: This is true, but we want to do this in the future and it should be a consideration in the design here.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+  uploads/
+harmonized/
+```
+
+Under `source` are data files such as raw read files received from the sequencing center, MRI DICOMs from the investigator, and so forth. The `uploads` directory contains files that are uploaded from the Data Tracker. `harmonized` contains files that were processed by the BIX team using things such as a standard alignment workflow or calling pipeline.
+
+### Issues with Study Buckets
+
+## Storage Backend Solution
+By allowing users to create "Storage Backends" that grant Dewrangle access to bucket resources in other accounts, we will be able to provide more flexibility in how organizations choose to store their data.
```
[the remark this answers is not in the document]
dankolbman: How about I replace Dewrangle and Data Tracker with 'the application'?

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+The Data Tracker relies on a data storage system to both save files that it receives and to account for other files that have been deposited externally. The current method that is used is adequate for a single tenant operating in a single AWS account, but needs to be re-thought for multiple tenants that  have more sophisticated storage needs.
+
+### Study Buckets
+
+Studies are central entities in the Data Tracker. They partition data that is collected and submitted by different research projects and investigators. For this reason, it was determined that each study would have its own bucket to segment it from other collections of data.
+
+#### Creation
+The Data Tracker automatically creates and configures buckets upon creation of a new study. These buckets are created in the same account as the IAM role that the API uses (Kids First Open Data). This account happens to be different from the role where the API actually resides (Kids First STRIDES). This is necessary as compute used to be contained within Kids First Open Data but was migrated to the STRIDES account. However, the data did not move and the codebase required that the IAM used was owned by the same account as the data.
```
gsantia, line 18: I know that this paragraph is trying to describe something complicated but it's pretty confusing

gsantia: should `This account happens to be different from the role where the API actually resides (Kids First STRIDES).` say `This IAM role happens to be different from the role where the API actually resides (Kids First STRIDES).`

?

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+Under `source` are data files such as raw read files received from the sequencing center, MRI DICOMs from the investigator, and so forth. The `uploads` directory contains files that are uploaded from the Data Tracker. `harmonized` contains files that were processed by the BIX team using things such as a standard alignment workflow or calling pipeline.
+
+### Issues with Study Buckets
+
+## Storage Backend Solution
+By allowing users to create "Storage Backends" that grant Dewrangle access to bucket resources in other accounts, we will be able to provide more flexibility in how organizations choose to store their data.
+
+## Default Storage
+![enter image description here](https://i.ibb.co/vBb1YsN/storage-layout.png)
+"Default Storage" will refer to the defaults for how a new organization and its studies will be stored. It's important that new organizations and studies be quickly bootstraped into a working state without unnecessary configuration. To achieve this, Dewrangle will provide a default workspace for documents to land. This will be achieved by providing a Storage Backend that will point to Dewrangle's application bucket.
```
gsantia, line 41:
```suggestion
![enter image description here](https://i.ibb.co/vBb1YsN/storage-layout.png)

"Default Storage" will refer to the defaults for how a new organization and its studies will be stored. It's important
that new organizations and studies be quickly bootstrapped into a working state without unnecessary
configuration. To achieve this, Dewrangle will provide a default workspace for documents to land. This will be
achieved by providing a Storage Backend that will point to Dewrangle's application bucket.
```
the formatting got weird

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+  SD_00000000/
+    source/
+    uploads/
+    harmonized/
+```
+
+## External Storage Backends
+
+![enter image description here](https://i.ibb.co/PTPMLnf/storage-layout.png)
+Under circumstances that an organization requires all of their uploads or data files to be under their control, an organization administrator may specify credentials and details for a bucket of their choosing in a different account.
```
gsantia, line 60:
```suggestion
![enter image description here](https://i.ibb.co/PTPMLnf/storage-layout.png)
When an organization requires all of their uploads or data files to be under their control, an organization administrator may specify credentials and details for a bucket of their choosing in a different account.
```
formatting

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+### Application Bucket
+
+It's proposed that Dewrangle maintain a single bucket in which it may stash data. This bucket will be intended to contain anything the Dewrangle may need to store such as configuration files, user avatars, and similar. Most notably, though, it will contain uploaded documents for each study. In this way, the single application bucket will deprecate the need for each study bucket to contain its uploads. However, it is also recognized that perhaps some organizations may wish their uploaded documents to reside within their own resource pool. For this case, we will allow the document bucket to be specified within a different account.
+The default solution for a new organization (and the existing Kids First organization) will be to store all documents within the application bucket. These documents will be kept as outlined above in the `New Layout` where the `studies/` prefix will be at the root of the bucket.
+
+### New Layout
+```
+studies/
+  SD_00000000/
+    source/
```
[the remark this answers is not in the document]
dankolbman: Yes, I think this is a good move to consider the uploads as a third set of data that is neither 'harmonized' or 'raw'. However, it is not absolutely necessary that the documents reside outside the source, we may still specify document backends that reside within any prefix, even source.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+    uploads/
+    harmonized/
+```
+
+## External Storage Backends
+
+![enter image description here](https://i.ibb.co/PTPMLnf/storage-layout.png)
```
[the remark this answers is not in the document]
dankolbman: The label of the diagram should really read 'bucket structure' as it's supposed to refer to the schema of the bucket rather than it's contents.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Use of Storage Backends
+
+A Storage Backend may be used in one of two ways: Either as a Document Storage Backend or a Data Storage Backend. Each method will require the same configuration of a Storage Backend and, in fact, a single Storage Backend may be used for both Documents and Data.
+
+#### Document Storage Backends
+
+Storage Backends used for documents will be read/write. Dewrangle will use these backends to save documents using a `uploads/` prefix under the root of the Storage Backend.
```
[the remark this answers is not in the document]
dankolbman: Yeah I think this is a good idea to namespace the root. We could do something like `dewrangle/uploads/` as a prefix prepended automatically to everything.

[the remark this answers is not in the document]
dankolbman: We can have `dewrangle` be it's own prefix because it we may want some other contents later in parallel with the uploads. Perhaps we have some extract configs directory or a prefix for holding intermediate results from ETL.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Data Storage Backends
+
+Storage Backends used for data will only utilize read and list operations. These backends will be used for file discovery and accounting only.
```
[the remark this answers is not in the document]
dankolbman: Yes, it could be up to the user providing the IAM user for the target account. For simplicity, the design here doesn't differentiate between the two, though. It seems too complex to have two different types of storage backends and flows to create them. It is much easier to have a single Storage Backend concept and allow them to be assigned to different tasks even if they may be slightly over-provisioned for some of those tasks.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Configuration
+To configure a storage backend, an organization administrator will be required to provide details about the AWS account where their target bucket resides and a user who may access it. An organization administrator will be able to configure this by navigating to the *Organization Settings* screen and choosing *Storages*. This will allow the administrator to add a new storage by first asking the user to create a new IAM user in their AWS account and assigning a specified policy. After the user has done this, they will be required to provide the AWS access and secret key pair so that Dewrangle may access the bucket. The final step will be to allow the user to specify the bucket and what prefix under it they would like this storage backend to use as the root. After confirming their choices, Dewrangle will attempt to access the bucket with the provided credentials and save the new storage backend if successful.
```
gsantia, line 67: i'm not sure what `The final step will be to allow the user to specify the bucket and what prefix under it they would like this storage backend to use as the root.` means

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Data Storage Backends
+
+Storage Backends used for data will only utilize read and list operations. These backends will be used for file discovery and accounting only.
+
+### Setting Organization Storage Backend Defaults
+
+If an organization administrator wishes all new studies to use a given backend for document uploads or data files, they may specify it in the organization settings. The administrator will have two defaults to configure: *Default Document Storage Backend* and *Default Data Storage Backend*. An administrator may use the same Storage Backend for both, but it will be up to them to prevent collisions.
+
+### Changing Storage Backends Per-Study
```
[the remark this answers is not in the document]
dankolbman: I believe I address the first question somewhere in here. The answer is that it is the responsibility of the user to make those decisions and contact us if they need help migrating old files.

For the second question, yes, the suggested method here would be create a new storage backend with a more specific prefix. This should not be encouraged though because a single storage backend will automatically prefix studies in its own way.

The uniqueness should probably be constrained to org+bucket+prefix. I don't see a reason to have the same org+bucket+prefix but using a different user, but I will have to think about it some more.

docs/rfcs/XXXX_storage_backends.md
dankolbman: Yeah this may not be accurate. Because studies have both a one-one for document storage and a one-many for data storage, django will put two reverse relations on this model.

docs/rfcs/XXXX_storage_backends.md
dankolbman: Because it is not necessary to get access. For instance, if we own the bucket we want to use for a storage backend and we have control over the application's IAM user, we are able to grant it access without having to create a new user or insert new credentials. We just need to leave these fields blank and the application will attempt to access the resource using the application's role.

docs/rfcs/XXXX_storage_backends.md
dankolbman: Yes I can see the appeal of this approach, however, we should encourage new users per-storage backend to be created and used. This ensures each user has minimal reach which limits attack surface in the case of being compromised. Consequently, it does have a trade off with being more difficult to cycle keys as each user would have to be updated.

docs/rfcs/XXXX_storage_backends.md
dankolbman: We can do that, I think we may even be able to retrieve it automatically if the credentials checkout.

docs/rfcs/XXXX_storage_backends.md
dankolbman: I imagine that reading from S3 is a lot slower than from the database. We'd also materialize these files in S3 so we'd have to make sure they are pretty well administered to prevent accidental exposure. Keeping them in the database provides two layers of security, one from the db connection auth, and one from the KMS.

docs/rfcs/XXXX_storage_backends.md
dankolbman: This is a field that may be set automatically by a periodic health check that will ensure the provided credentials give the right level of access.

docs/rfcs/XXXX_storage_backends.md
dankolbman: Does the user flow where the user is asked to provide bucket info make more sense? Prefixes are like folders in S3, so the user may optionally provide a prefix inside the bucket where all the data where reside. It's like git looking for `.git` and only being concerned about whatever files lie below it in the file system.

gsantia: yeah that makes sense

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+It may be desirable for a study administrator to use different Storage Backends for different studies. For this, studies will be able to specify one Storage Backend for documents and one or more Storage Backends for data files. 
+
+#### Kids First Example
+
+In Kids First, each study keeps its data in different buckets including documents. If we were to use the Storage Backend approach for Kids First, we could do so by first creating a new Storage Backend for each bucket and then updating each study with the appropriate Storage Backend.
+If we would like to stop distributing uploaded documents to each study's bucket, we could simply update each study's Document Storage Backend to use the Default Storage Backend. This would result in all uploaded documents going to the application's bucket while all data operations occurring on each individual study's Data Storage Backend.
```
gsantia, line 92:
```suggestion
If we would like to stop distributing uploaded documents to each study's bucket, we could simply update each study's Document Storage Backend to use the Default Storage Backend. This would result in all uploaded documents going to the application's bucket with all data operations occurring on each individual study's Data Storage Backend.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,248 @@
...
+
+### StorageBackend Model
+
+The storage backend model
+
+```graphql
+type StorageBackend {
+  id: ID!
+  name: String!
+  bucket: String
```
gsantia, line 150: why would `id` and `name` be the only things that are required?

docs/rfcs/XXXX_storage_backends.md
dankolbman: Should be able to get the account id, alias, and arn which will yield the user's name.

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+Study buckets are restricted to the same account as the IAM profile assigned to the application's compute instance.
+The IAM profile contains a large surface of permissions needed to create and modify buckets.
+
+## Storage Backend Solution
+By allowing users to create "Storage Backends" that grant the application access to bucket resources in other accounts, we will be able to provide more flexibility in how organizations choose to store their data.
+
+## Default Storage
+![enter image description here](https://i.ibb.co/vBb1YsN/storage-layout.png)
+"Default Storage" will refer to the defaults for how a new organization and its studies will be stored. It's important that new organizations and studies be quickly bootstraped into a working state without unnecessary configuration. To achieve this, the application will provide a default workspace for documents to land. This will be achieved by providing a Storage Backend that will point to the application's application bucket.
```
shanewilson, line 45: > It's important that new organizations and studies be quickly bootstraped into a working state without unnecessary configuration

just curious but why is it important for studies to be bootstrapped quickly? won't that lead to files being uploaded into the wrong bucket once the study is properly setup? and then they will need to migrate the data from the default bucket into their bucket - or re-upload everything

shanewilson:
Are there any example of companies currently doing this sort of thing that we can use to compare?

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Status:** Approved, waiting for implementation
+
+
+## Background and Existing Solution
+
+![enter image description here](https://i.ibb.co/C7bzPNZ/storage-layout.png)
+> A simple representation of how the application and its resources are currently allocated.
+
+The the application relies on a data storage system to both save files that it receives and to account for other files that have been deposited externally. The current method that is used is adequate for a single tenant operating in a single AWS account, but needs to be re-thought for multiple tenants that  have more sophisticated storage needs.
```
gsantia, line 11:
```suggestion
The application relies on a data storage system to both save files that it receives and to account for other files that have been deposited externally. The current method is adequate for a single tenant operating in a single AWS account, but needs to be re-thought for multiple tenants that have more sophisticated storage needs.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Study Buckets
+
+Studies are central entities in the the application. They partition data that is collected and submitted by different research projects and investigators. For this reason, it was determined that each study would have its own bucket to segment it from other collections of data.
```
gsantia, line 15:
```suggestion
Studies are central entities in the application. They partition data that is collected and submitted by different research projects and investigators. For this reason, it was determined that each study would have its own bucket to segment it from other collections of data.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Creation
+
+The the application automatically creates and configures buckets upon creation of a new study. These buckets are created in the same account as the IAM role that the API uses (Kids First Open Data). This account happens to be different from the account where the API actually resides (Kids First STRIDES). This is necessary as compute used to be contained within Kids First Open Data but was migrated to the STRIDES account. However, the data did not move and the codebase required that the IAM used was owned by the same account as the data.
```
gsantia, line 19:
```suggestion
The application automatically creates and configures buckets upon creation of a new study. These buckets are created in the same account as the IAM role that the API uses (Kids First Open Data). This account is different from the account where the API actually resides (Kids First STRIDES). This is necessary as compute used to be contained within Kids First Open Data, but was migrated to the STRIDES account. However, the data did not move and the codebase required that the IAM used was owned by the same account as the data.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Configuration
+
+Each study bucket needs to be configured with the correct security policies, encryption, replication, logging, and so on. For this reason, the bucket creation process was automated and included as part of the the application's service offerings. Upon creation, the the application will configure these additional settings ensuring all buckets are setup correctly.
```
gsantia, line 23:
```suggestion
Each study bucket needs to be configured with the correct security policies, encryption, replication, logging, and so on. For this reason, the bucket creation process was automated and included as part of the the application's service offerings. Upon creation, the application will configure these additional settings ensuring all buckets are set up correctly.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+![enter image description here](https://i.ibb.co/vBb1YsN/storage-layout.png)
+"Default Storage" will refer to the defaults for how a new organization and its studies will be stored. It's important that new organizations and studies be quickly bootstraped into a working state without unnecessary configuration. To achieve this, the application will provide a default workspace for documents to land. This will be achieved by providing a Storage Backend that will point to the application's application bucket.
+
+### Application Bucket
+
+It's proposed that the application maintain a single bucket in which it may stash data. This bucket will be intended to contain anything the the application may need to store such as configuration files, user avatars, and similar. Most notably, though, it will contain uploaded documents for each study. In this way, the single application bucket will deprecate the need for each study bucket to contain its uploads. However, it is also recognized that perhaps some organizations may wish their uploaded documents to reside within their own resource pool. For this case, we will allow the document bucket to be specified within a different account.
```
gsantia, line 49:
```suggestion
It's proposed that the application maintain a single bucket in which it may stash data. This bucket will be intended to contain anything the application may need to store - such as configuration files, user avatars, etc. Most notably, though, it will contain uploaded documents for each study. In this way, the single application bucket will deprecate the need for each study bucket to contain its uploads. However, it is also recognized that perhaps some organizations may wish their uploaded documents to reside within their own resource pool. For this case, we will allow the document bucket to be specified within a different account.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+The default solution for a new organization (and the existing Kids First organization) will be to store all documents within the application bucket. These documents will be kept as outlined above in the `New Layout` where the `studies/` prefix will be at the root of the bucket.
```
gsantia, line 50:
```suggestion
The default solution for a new organization (and the existing Kids First organization) will be to store all documents within the application bucket. These documents will be kept as outlined below in the `New Layout` where the `studies/` prefix will be at the root of the bucket.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+		SD_00000000/
+			source/
+			uploads/
+			harmonized/
+```
+
+## External Storage Backends
+
+![enter image description here](https://i.ibb.co/PTPMLnf/storage-layout.png)
+When/ an organization requires all of their uploads or data files to be under their control, an organization administrator may specify credentials and details for a bucket of their choosing in a different account.
```
gsantia, line 67:
```suggestion
When/if an organization requires all of their uploads or data files to be under their control, an organization administrator may specify credentials and details for a bucket of their choosing in a different account.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Use of Storage Backends
+
+A Storage Backend may be used in one of two ways: Either as a Document Storage Backend or a Data Storage Backend. Each method will require the same configuration of a Storage Backend and, in fact, a single Storage Backend may be used for both Documents and Data.
```
gsantia, line 79:
```suggestion
A Storage Backend may be used in one of two ways: a Document Storage Backend or a Data Storage Backend. Each method will require the same configuration of a Storage Backend and, in fact, a single Storage Backend may be used for both Documents and Data.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Document Storage Backends
+
+Storage Backends used for documents will be read/write. the application will use these backends to save documents using a `dewrangle/uploads/` prefix under the root of the Storage Backend.
```
gsantia, line 83:
```suggestion
Storage Backends used for documents will be read/write. The application will use these backends to save documents using a `dewrangle/uploads/` prefix under the root of the Storage Backend.
```

docs/rfcs/XXXX_storage_backends.md
```diff
@@ -0,0 +1,288 @@
...
+It may be desirable for a study administrator to use different Storage Backends for different studies. For this, studies will be able to specify one Storage Backend for documents and one or more Storage Backends for data files. 
+
+#### Kids First Example
+
+In Kids First, each study keeps its data in different buckets including documents. If we were to use the Storage Backend approach for Kids First, we could do so by first creating a new Storage Backend for each bucket and then updating each study with the appropriate Storage Backend.
+If we would like to stop distributing uploaded documents to each study's bucket, we could simply update each study's Document Storage Backend to use the Default Storage Backend. This would result in all uploaded documents going to the application's bucket with all data operations occurring on each individual study's Data Storage Backend.
+
+### Management of External Buckets
+
+the application will not be responsible for configuration of the external buckets. This responsibility resides with the organization administrator and whoever controls their resources.
```
gsantia, line 104:
```suggestion
The application will not be responsible for configuration of the external buckets. This responsibility resides with the organization administrator and whoever controls their resources.
```

gsantia (commented):
minor spelling/grammar stuff

docs/rfcs/XXXX_storage_backends.md
dankolbman: Yes, the files would flow to potentially the wrong location if the study was setup without the storage backend ready to go.
The ability to quickly setup and start uploading to a study is more about stream-lining the process for the administrative user. In our case, the user creating the study is not the one with the know-how to setup and configure the storage backend. This leads to a blocker in the time-to-first-upload process.

However, another benefit of the bootstrapping was also that we could ensure each study bucket was being configured identically. Now that we are moving to a user-managed model for storage backends, this is less of a benefit.

Because using the 'default storage backend' will be hostile to users who will need to move to a new backend, perhaps we should not allow the use of a default until we also allow seamless migration. Maybe we start with a more restrictive feature set and aim to add this 'default' later as the next iteration:
1) Require users to select or create a new storage backend when creating their study
2) Allow users to select a 'default storage backend' when creating their study and enable seamless migration of data when the user changes their study's backend after it's been created.

dankolbman:
@shanewilson There isn't a strict requirement that I know of that mandates we keep files as files. The main benefit of keeping the file contents in the database is that we don't need to worry about coordinating different buckets in different accounts? This sounds like a big plus, but won't we still need to coordinate buckets and accounts if there's a requirement to copy/dump files into destination locations anyway?

It's not a lot of work to write a translation layer between the database and actual files, but consider that we'd essentially null out any tools that operate on a filesystem. We would greatly limit data consumers by requiring them to use our api to retrieve documents. There's also not the same 'feeling' and sense of ownership and security that comes with seeing material files in a bucket somewhere, which I believe is pretty important to the user.

znatty22:
@shanewilson Adding my two cents on to dan's point. Perhaps this shouldn't be taken into consideration but as a developer (or maybe even analyst?), I think it's useful to have the ability to go into S3 and browse actual files and download them if I need to for any reason.

PR #726, opened 2021-08-18
🔊 Configure TaskLogger directly

dankolbman:
Configures the TaskLogger directly in the top logging configuration to ensure that task logs are produced and stored.

This is not desired long-term as it will send these logs directly to stdout which will get captured in the cloudwatch logs for the application. This will result in a lot of doubly-stored and unnecessary cruft going into places like cloudwatch and s3.

### DIFF
--- a/creator/settings/production.py
+++ b/creator/settings/production.py
@@ -295,7 +295,7 @@ LOGGING = {
         },
     },
     "loggers": {
-        "TaskLogger": {},
+        "TaskLogger": {"handlers": ["task"], "level": "INFO"},
         "graphql.execution.utils": {
             "handlers": ["command"],
             "level": "CRITICAL",


creator/settings/production.py
```diff
@@ -295,15 +295,14 @@ def traces_sampler(sampling_context):
...
-        "TaskLogger": {},
+        "TaskLogger": {"handlers": ["task"], "level": "INFO"},
         "graphql.execution.utils": {
             "handlers": ["command"],
             "level": "CRITICAL",
         },
         "creator": {"handlers": ["task"]},
         "creator.management": {"handlers": ["command"], "level": "INFO"},
         "rq.worker": {"handlers": ["rq_console"]},
-        "creator.decorators": {"handlers": ["task"]},
```
znatty22, line 306: Why remove this? Don't you still want some of those log statements in the task decorator app code to show up in cloudwatch?

znatty22 (approved):
🔔

PR #730, opened 2021-08-19
✨ Add endpoint to download template versions

### LINKED ISSUE #727: Download any templates as package endpoint
We need a new download endpoint for downloading any set of templates from the template admin view on Data Tracker. 

Currently templates can only be downloaded from the study's data template view. This works well for users that work within the context of a study but not for others (e.g. admins) that want to download templates from outside the study context (maybe they want to download another org's template and modify that to create a new one for their org).

**Task**
Implement new endpoint with similar functionality as the download study templates endpoint

```
/downloads/templates?file_format=[excel|zip]&template_versions=<comma separated list of TemplateVersion.id>
```

**Code**
Changes need to be made in these files

`creator/urls.py`
`creator/data_templates/views.py`

Please feel free to ask questions if anything is confusing or missing from this task description!

gsantia:
Closes #727 

Adds a download endpoint from the template admin view on DataTracker to download any set of templates. Right now only templates within a single study can be downloaded in one request, which is not convenient for admins.

### DIFF
--- a/creator/data_templates/packaging.py
+++ b/creator/data_templates/packaging.py
@@ -11,14 +11,12 @@ from creator.data_templates.models import TemplateVersion
 
 MAX_SHEET_NAME_LEN = 31
 
-PACKAGE_INFO = (
-    """
+PACKAGE_INFO = """
     The Fields file describes all of the fields or columns in the template
     (name, data type, required or not, etc.) while the blank Template file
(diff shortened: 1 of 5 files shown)


znatty22 (approved):
👍

dankolbman (approved):
# 🥞

PR #738, opened 2021-08-31
💬 Add user to template event description

### LINKED ISSUE #737: Incorporate the user into template event description
We should incorporate the user taking an action on the template in the template event description just like the other file events:

image

*Code Changes*
`creator/templates/signals.py`
May be other changes needed to capture the user taking action .. needs some further research

gsantia:
Closes #737.

Incorporate the user into the template event description like in the other file events. As far as I can tell, it works exactly the same here.


znatty22 (approved):
👍

PR #739, opened 2021-09-01
🐛 Delete templates in faketemplates cmd if flag is given

znatty22:
Templates were not being deleted when user passed --delete option


gsantia (approved):
🪑
any idea why `store_true` and `store_false` are named that way? kinda confusing

dankolbman (approved):
# ⛵️

PR #741, opened 2021-09-10
♻️ Restore old general file types and add new ones

znatty22:
Users want to categorize their file regardless of whether it conforms to a template or not. The general file types provide this

### Changes 

- Bring back the old general file types and add new ones
- Migrate files marked as "Other" to a more specific file type. Use file name to guess file type
- Fix the tests that break as a result of changes

Migration was tested locally on a copy of production data from 9/14/2021


creator/files/migrations/0027_restore_old_file_types.py
```diff
@@ -0,0 +1,80 @@
...
+    - Attempt to change the file type of all files marked as OTH to something
+    more specific
+"""
+
+from django.db import migrations, models
+
+
+def set_file_type(apps, schema_editor):
+    """
+    Try our best to set files with file_type OTH to a more specific file type
```
dankolbman, line 15: Are there any risks to consider if a file gets misclassified?

gsantia (approved):
🎣

creator/files/migrations/0027_restore_old_file_types.py
znatty22: Yea I'm not sure. Probably a good question for @chris-s-friedman. When I ran the mapping function on prod data from Friday it seemed to work fairly well. I could have @chris-s-friedman take a look at the results of the data migration so we can be sure nothing gets misclassified?

chris-s-friedman: > I could have @chris-s-friedman take a look at the results of the data migration so we can be sure nothing gets misclassified? 

I can do that :)

gsantia (approved):

dankolbman (approved):
# 🫑

znatty22:
Waiting on Chris F to send me file type corrections so they can be incorporated into the data migration

PR #742, opened 2021-09-13
♻️ Refactor validation to use templates

### LINKED ISSUE #723: Refactor validation to use templates
The current validation functionality relies on a lot of hardcoded functionality that is specific to Kids First templates and extract configs. We need to refactor validation such that:

**Refactor Validation Task**
- [x] Remove dependence on hardcoded extract configs (`validation_run.clean_and_map`)
- [x] Replace `clean_and_map` with a function that maps the source data columns to template keys

**Future Work**
- [ ] Replace default basic validation rules from ingest library with better rules that incorporate S3 files

znatty22:
Closes #723 

Removes dependence on hardcoded ExtractConfigs to run validation. Validation now uses templates to map source file columns to standard concepts rather than ExtractConfigs.

### DIFF
--- a/creator/ingest_runs/tasks/validation_run.py
+++ b/creator/ingest_runs/tasks/validation_run.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+from typing import List, Dict, Optional
 import os
 import jsonpickle
 import logging
@@ -19,7 +21,9 @@ from kf_lib_data_ingest.validation.reporting.markdown import (
 
 from creator.decorators import task
 from creator.ingest_runs.models import ValidationRun, ValidationResultset
+from creator.studies.models import Study
 from creator.files.models import Version
+from creator.data_templates.models import TemplateVersion
 from creator.analyses.file_types import FILE_TYPES
 from creator.analyses.analyzer import extract_data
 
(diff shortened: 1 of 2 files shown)


creator/ingest_runs/tasks/validation_run.py
```diff
@@ -28,51 +32,108 @@
...
+    # it does, could mean inconsistent data in the source file
+    reverse_mapping = defaultdict(set)
+    for src, key in columns_to_keys.items():
+        reverse_mapping[key].add(src)
+    duplicates = {
+        key: src_cols
+        for key, src_cols in reverse_mapping.items()
+        if len(src_cols) > 2
+    }
+    if duplicates:
```
dankolbman, line 79: The above logic may be more clearly expressed as:
```
duplicates = Counter(columns_to_keys.values())
if max(duplicates.values()) > 2:
  logger.warning()
```

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -28,51 +32,108 @@
...
-            "not be included in validation."
+            "Found multiple columns in source file(s) that map to the same "
+            f"template key:\n{pformat(duplicates)}"
         )
-        raise ExtractConfigFileDoesNotExist
+
+    return columns_to_keys
+
+
+def map_column(in_col: str, mapper: Dict[str, str]) -> str:
```
dankolbman, line 88: The return type here is `Optional[str]`

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -28,51 +32,108 @@
...
-        raise type(e)(er_msg).with_traceback(sys.exc_info()[2])
+        raise ExtractDataError from e
 
-    logger.info(
-        f"Applied extract config {os.path.split(ec)[-1]} for file "
-        f"{version.root_file.file_type}, {version.kf_id}"
-    )
+    # Try to map as many of the input columns to template keys
+    mapped_cols = {c: map_column(c, mapper) for c in df.columns}
+    mapped_df = df.rename(columns=mapped_cols, errors="ignore")
```
dankolbman, line 125: I think you can condense the two above lines to:
```
mapped_df = df.rename(columns=map_column, errors="ignore")
```

znatty22: The `map_column` function still needs the `mapper` input and I think `rename` only lets you pass in a func

creator/ingest_runs/tasks/validation_run.py
gsantia: why is it `len(src_cols) > 2`? should it be `len(src_cols) > 1`?

if we want to keep the log message the same, could do

```
vals = set()
duplicates = {}
for key, val in columns_to_keys.items():
    if val not in vals:
        vals.add(val)
    else:
        duplicates[key] = val
if duplicates:
    logger.warning(
        "Found multiple columns in source file(s) that map to the same "
        f"template key:\n{pformat(duplicates)}"
    )
```

gsantia: why is it `len(src_cols) > 2`?

znatty22: I like this approach for the case where I don't care about displaying src cols with the same mapped template key

znatty22: > why is it len(src_cols) > 2

@gsantia  Bc every template column gets mapped to a template key and then I also added a mapping from every template key to template key. I did this bc I want to be able to extract columns from the src file which match template columns or template keys

dankolbman (approved):
# 🧢

creator/ingest_runs/tasks/validation_run.py
```diff
@@ -83,31 +148,78 @@ def validate_file_versions(validation_run: ValidationRun) -> dict:
...
+        raise TemplateVersion.DoesNotExist(
+            "Unable to run validation without templates. This study "
+            f"{study.pk} does not have any templates assigned to it"
+        )
+
+    # Generate mapping from template columns to template keys
+    mapper = generate_mapper(template_versions)
+    if not mapper:
+        raise ValueError(
+            "Unabled to run validation. Study {study.pk} templates do not "
```
gsantia, line 164:
```suggestion
            f"Unable to run validation. Study {study.pk} templates do not "
```

znatty22: Thank you :)

tests/ingest_runs/test_validation.py
```diff
@@ -219,56 +292,98 @@ def test_persist_results(db, tmpdir, settings, data_review):
...
+
+def test_generate_mapper(db, mocker):
+    """
+    Test validation_run.generate_mapper
+    """
+    class MockTemplate:
+        def __init__(self, fields):
+            self.field_definitions = {"fields": fields}
+    fields = [
+        {f"key": f"key{i}", f"label": f"label{i}"}
```
gsantia, line 338:
```suggestion
        {"key": f"key{i}", "label": f"label{i}"}
```

gsantia (approved):

PR #744, opened 2021-09-17
✨ Mutation to create Flatfile settings from templates

### LINKED ISSUE #745: Flatfile Settings Mutation
The frontend is integrating the Flatfile Portal import button so that users may do inline file mapping and editing. This button needs to be configured with the kind of information stored in our data template fields. 

We need a mutation that is able to produce the Flatfile settings JSON from template fields.

znatty22:
Closes #745 

A mutation to support frontend https://github.com/kids-first/kf-ui-data-tracker/issues/1111

image

### DIFF
--- a/creator/data_templates/flatfile_settings_schema.py
+++ b/creator/data_templates/flatfile_settings_schema.py
@@ -1,0 +1,237 @@
+"""
+Schemas for Flatfile settings JSON which is required
+as input to the Flatfile Portal import button in the Data Tracker frontend app.
+
+Flatfile settings are composed from the field definitions of
+template versions. In most cases, the settings object will be
+built from a list of study's templates.
+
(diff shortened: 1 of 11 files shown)


creator/data_templates/flatfile_settings_schema.py
```diff
@@ -0,0 +1,237 @@
...
+
+        See https://flatfile.com/developers/javascript/fields/
+        """
+        out_data = {}
+
+        # Field key and label
+        tlabel = template_field.get("label")
+        out_data["label"] = tlabel
+        # Use label as the key
+        out_data["key"] = tlabel
```
gsantia, line 99: why are these the same?

gsantia (approved):
🗃

needs a rebase

PR #746, opened 2021-09-17
🐛 Fix processing of missing value list

### LINKED ISSUE #747: Missing values list is not converted to str
The missing values attribute in a template field has a list of values. When the templates fields are serialized before download, any list types should be converted to delimited strings. This works for the accepted values list but not for the missing values list.

znatty22:
Closes #747


dankolbman (approved):
# 🧄

gsantia (approved):

Issue #749, opened 2021-09-21
Replace file readers with ingest library's read_df


znatty22:
We need to replace the individual file readers in `creator.analyses.analyzer` with the ingest library's `read_df` method.

The ingest library's `read_df` is already able to read all of the file formats covered by the individual read functions in `creator.analyses.analyzer` so these read functions are redundant. And more importantly, the ingest library's `read_df` automatically detects the encoding and properly decodes the file content. Right now the `CSVDictReader` assumes utf-8 encoding which can be wrong, resulting in the BOM left in the file. This has resulted in silent validation failures since the BOM may be part of the first file column and then is unrecognized by the validator.

Since this change would affect critical functionality in study creator, we can simple use `read_df` in the validation code to start.

PR #750, opened 2021-09-21
🐛 Properly decode file content before validation

znatty22:
Workaround for #749


gsantia:
codecov patch check is failing, does that matter?

creator/files/models.py
```diff
@@ -369,6 +370,22 @@ def extract_config_path(self):
...
         return config_path
 
+    def set_storage(self):
+        """
+        Set storage location for study bucket if using S3 backend
+        """
+        s3_storage = "django_s3_storage.storage.S3Storage"
+        if settings.DEFAULT_FILE_STORAGE == s3_storage:
+            if self.study is not None:
+                study = self.study
```
gsantia, line 380: this line isn't being covered in tests

creator/files/models.py
```diff
@@ -369,6 +370,22 @@ def extract_config_path(self):
...
+            elif self.root_file is not None:
+                study = self.root_file.study
+            else:
+                raise Study.DoesNotExist(
+                    f"Version {self} must be part of a study."
```
gsantia, line 385: for consistency could import the `version_display` function in this file then do
```
                    f"Version {version_display(self)} must be part of a study."
```

Issue #743, opened 2021-09-14
Experience delayed response on calling create new version mutation


XuTheBunny:
Experience errors when uploading a new version
https://data-tracker.kidsfirstdrc.org/study/SD_Z6MWD3H0/documents/SF_JF6YJRDK
![image](https://user-images.githubusercontent.com/32206137/133335578-070b0c84-81ed-4f56-b1b8-3865cfdc0359.png)
<img width="755" alt="Screen Shot 2021-09-14 at 1 59 16 PM" src="https://user-images.githubusercontent.com/32206137/133335579-9ed9cac0-7681-4c94-9256-85aa9e3e2966.png">

Noticed that the when the create new version mutation was called it took longer to get the response back, and a long waiting can lead to the call fail.

znatty22:
@XuTheBunny Do you know how large the file was? Wondering if that could be a cause

XuTheBunny:
The file is 18MB and has 54517 rows

znatty22:
Ahh yea that does seem on the larger side..I have one idea on where the bottleneck _could_ be..May possibly be in the `analyze_version` method bc that iterates over all of the columns and values to tally up distinct values. I remember when we tried this approach in the ingest library when reporting counts, we ran into performance issues. We found it was better to use pandas for this

znatty22:
Hmm I have one idea for where the bottleneck could be. The `analyze_version` method gets called in the upload version mutation and it iterates over all rows to tally up distinct values in each column. When we tried this approach in the ingest library, it became a bottleneck for large files. We found that pandas worked better for reporting unique counts

znatty22:
Closing because we are not going address this now. The cause was due to intermittent poor network connections. Increasing the Gunicorn web server timeout would have just been a brittle bandaid so decided to leave it as is.

PR #752, opened 2021-09-22
🐛 Change gunicorn timeout to 60 seconds

gsantia:
Partially addresses #743 

Gunicorn defaults to timeout after 30 seconds. As a work-around for slow uploads, change this to be 60 seconds instead.


znatty22:
@gsantia After some thought, let's leave this at 30 seconds since the duration of upload and resulting timeout is so dependent on individual network latency. I think increasing the timeout would just be a temporary bandaid. There are other approaches to this which we can try in the future that would be much better longer term solutions.

PR #753, opened 2021-09-22
♻️ Adjust validation rules and report

### LINKED ISSUE #754: Remove non-relevant sections from validation report
We need to remove certain sections of the validation report that aren't relevant or are obsolete. 

- Remove Count Tests (we don't have expected counts to test against)
- Remove Gap Tests (this is an internal / obscure test that most users don't understand)
- Remove Attribute Value Tests, e.g. enum values, range checks (These are hardcoded to the KF concepts and won't make sense when a study has its own templates)

### LINKED ISSUE #755: Replace default relationship validation rules with better rules
The validation library has its own set of validation rules that it uses to do relationship validation, but most of these rules are either hardcoded to KF relationships or don't have enough specificity (no representation of specimen vs aliquot, no differentiation between source sequencing files and harmonized genomic files on S3). 

The validation library allows you to pass in your own relationship rules. We should pass in a new set of relationship validation rules that are better/more specific than the defaults.

znatty22:
Closes #754 
Closes #755 

image


gsantia (approved):
📖

PR #761, opened 2021-09-27
🔊 Configure release task logging directly

znatty22:
Configures the logging for release tasks/jobs so that log statements make it to stdout. This is might not the long term solution since we don't necessarily want to have to configure logging for each individual module. But it is an important fix that will help us debug release issues until we figure out the longer term solution for API logging.


dankolbman (approved):
# 🍀

PR #763, opened 2021-09-28
♻️ Omit Content-Length header in download validation report

### LINKED ISSUE #762: Fix or remove content length in download validation report endpoint
There seems to be an issue with how django-s3-storage determines a files size and passes that to S3 when uploading a file. The validation report does not render correctly because the browser gets the incorrect `Content-Length` from the download endpoint which sources the size from the Django S3Storage backend. 

We should either omit the `Content-Length` header and let the client determine the size, or add a new size attribute to the `ValidationResultset` model to capture the file size when the report content is uploaded. Then we would manually set the `Content-Length` header with that file size in the download endpoint

znatty22:
Closes #762 

Tested in deployed dev and locally.

### DIFF
--- a/creator/ingest_runs/tasks/validation_run.py
+++ b/creator/ingest_runs/tasks/validation_run.py
@@ -324,9 +324,10 @@ def upload_validation_files(
         # Filename: validation_<report or results>_<data_review.kf_id>.<ext>
         fname, ext = os.path.splitext(fname)
         fname = f"{fname}_{resultset.data_review.kf_id.lower()}{ext}"
-        file_field.save(fname, ContentFile(content))
+        f = ContentFile(content)
+        file_field.save(fname, f)
 
-        logger.info(f"Wrote {file_field.name}")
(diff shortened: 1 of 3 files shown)


dankolbman:
This is an interesting bug. Any ideas why the s3-storages may be saving the wrong content length to s3?

znatty22:
@dankolbman  
> This is an interesting bug. Any ideas why the s3-storages may be saving the wrong content length to s3?

No, I didn't look at the src code closely. I want to eventually but I gotta get to some other stuff first

dankolbman (approved):
@znatty22 I looked into the changelog for django-s3-storage and I didn't see anything interesting that may solve this issue. If you look into it further, you may have to file a new issue.

gsantia (approved):

znatty22:
@dankolbman I bet it has to do with this block here:
https://github.com/etianen/django-s3-storage/blob/2fd1b66e34247346de1577cea78fe1340446cb68/django_s3_storage/storage.py#L308

PR #764, opened 2021-09-28
🐛 Add organization to newly invited users

### LINKED ISSUE #756: Investigate why invited users do not have an organization
There is an issue either in study creator, data tracker, or both with new user invitations. 

Here is the flow and where the issue happens:
1. Admin clicks Invite User button
2. Admin fill out form, add user to a study, assign a role
3. Admin click send invite
4. New user clicks link in email or logs into Data Tracker with the email they signed up with
5. User is not able to see anything in Data Tracker

The reason why this happens is, the new user has been added to a study, but there is some issue with the user's organization. Either they have not been added to an organization or the organization is not able to be resolved by the frontend. You can confirm this by 1. Go to  Admin > Users > 2. Search for the user by email 3. They have no organizations in their list.

Someone needs to investigate what is happening, determine whether its a backend or frontend issue, and potentially make a PR to fix it.

gsantia:
Hopefully closes #756


dankolbman (approved):
# 🦩

znatty22:
@gsantia Can you make one change to this PR? I need the get one referral token query to be unprotected by authorization checks bc otherwise the frontend can't query for the token in the new user registration flow

```
creator/referral_tokens/schema.py

Remove these lines:
       if not (user.has_perm("referral_tokens.view_referraltoken")):
            raise GraphQLError("Not allowed")
```

znatty22:
@dankolbman 

Yea it would be better to keep this query protected. The reason I wanted to query for the referral token is to get the new user’s email associated with the token so I could make sure the logged in user is the intended recipient of the token before calling the exchange token mutation. What was happening before is, I would be logged in as say nat1@email.com, send an invite to nat2@email.com, and then when I visit the invite url, nat1@email.com would end up claiming the token instead of nat2@email.com. I know that's sort of a rare scenario and could be solved by logging out or using a private window but I think we should be able to handle it in the app/api. 

So maybe a better way to handle this would be to let the exchange mutation be called with whoever is logged in and then in the mutation logic check to make sure that the authenticated user’s email matches the email in the referral token.

znatty22:
@gsantia Can you add this to the exchange token mutation after the referral_token.is_valid check?

```
        if user.email != referral_token.email:
            raise GraphQLError(
                f"{user.email} cannot use this referral link. Must be used by "
                f"the invited user"
            )
```

PR #767, opened 2021-09-29
🐛  Fix writing of job logs in local development

### LINKED ISSUE #766: Job logs are not written in local development
Job logs are not written to the local file system during development, making it difficult to develop and test frontend features that make use of the job logs. This because Django raises a `SuspiciousFileOperation` exception when the API tries to write logs to disk. It is likely due to a misconfiguration of the upload directory for job logs.

znatty22:
Closes #766 

Django raises a `SuspiciousFileOperation` exception if the API tries to write into the BASE_DIR. We needed to move the local logs directory outside base application directory

### DIFF
--- a/creator/jobs/models.py
+++ b/creator/jobs/models.py
@@ -87,7 +87,11 @@ def _get_upload_directory(instance, filename):
         prefix = f"{settings.LOG_DIR}/{filename}"
         return prefix
     else:
-        return os.path.join(settings.BASE_DIR, settings.LOG_DIR, filename)
+        # If not writing to S3, remove the date based folder hierarchy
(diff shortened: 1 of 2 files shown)


dankolbman (approved):
# 🐉

Had this issue crop up somewhere else, I think it was because django suddenly didn't like how there were path names in some places.

znatty22:
@dankolbman 
>Had this issue crop up somewhere else, I think it was because django suddenly didn't like how there were path names in some places.

The path names definitely causes that suspicious exception but I stripped those out and it was still raising the exception. It was only after I removed the BASE_DIR that it worked.

PR #768, opened 2021-09-30
🐛 Fix API logging

### LINKED ISSUE #760: Fix API logging
There still seems to be an issue with API logging. We're not sure if its specifically with the async tasks or with the API as a whole. This PR https://github.com/kids-first/kf-api-study-creator/pull/726 was an attempt to fix the task logging but it may not have worked. We need to investigate this issue as it is very difficult to debug production release coordinator issues without proper logging.

znatty22:
Closes #760 

Tested locally and in deployed dev. Now we can see log messages from all loggers under the parent creator logger in the console and the TaskLogger's stream

### DIFF
--- a/creator/releases/models.py
+++ b/creator/releases/models.py
@@ -138,7 +138,7 @@ class Release(models.Model):
     @transition(field=state, source="running", target="staged")
     def staged(self):
         """ The release has been staged """
-        return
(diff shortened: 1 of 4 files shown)


gsantia (approved):
did you mean to not change the `testing.py` settings file?

gsantia (approved):

PR #774, opened 2021-09-30
🔥 Remove sync job with old release coordinator

### LINKED ISSUE #773: Phase out sync job with old release coordinator
We can stop the periodic job which pulls in releases from the old release coordinator since that service is deprecated and all releases have been synced. The job is always failing and cluttering up logs with the stack trace anyway:

image

znatty22:
Closes #773


dankolbman (approved):
# 🐕‍🦺

PR #776, opened 2021-10-04
✨ Capture time of release termination 

### LINKED ISSUE #765: Capture time of release termination
It would be very helpful to capture the time that a release terminated whether it completed, failed, or canceled and then display that on the release history table (similar to Started At). 

Right now it is difficult to search cloudwatch logs for issues because I have to do some hunting to figure out when the release failed. Sure, I could scroll to the bottom of the release log and translate the UTC timestamp to the local timezone displayed in cloudwatch logs. But who wants to do that every time.

gsantia:
Closes #765 

Capture the time that a release terminated, including when it fails, completes, or cancels.


znatty22:
@gsantia You branch is failing deployment atm because its trying to apply your 0005_update_times.py migration except that a one of your previous migrations (probably named differently in an old commit) already added the ended_at column. I'm going to fix it in deployed dev so that this branch can deploy successfully and then I can show you how we normally deal with this issue next week. Just FYI for now

PR #789, opened 2021-10-26
✨ Add short_code, domain, and program to Study

### LINKED ISSUE #788: Add code, domain, and program to Study
We need to add a new fields to the Study model to support new search facets on the Portal. These fields are being added to the Dataservice's Study model as well in https://github.com/kids-first/kf-api-dataservice/issues/547. 

### short_code
- A short alphanumeric string that identifies the study in a succinct human friendly way
- Min length: 3
- Max length: 6
- Must be an alphanumeric string
- Must have unique constraint
- Not nullable

#### Migration
Since this field is required we'll need to do a multi-step migration where we apply the schema migration without the required constraint, do the data migration to fill in `short_code`, and then apply the required constraint. The data migration should use the sheet attached.

### domain
- The category of disease being studied
- enumeration: Cancer, Birth Defect, Cancer and Birth Defect, Other
- nullable
- default=Unknown

#### Migration
The data migration should use the sheet attached

### program
- The administrative organization or group of the study
- nullable
- default=Use organization name

#### Migration
The data migration should use the sheet attached

## Deployment
We need to coordinate deployments with the Dataservice deployments for https://github.com/kids-first/kf-api-dataservice/pull/550. Basically don't deploy this feature into an environment until the Dataservice side has been deployed in that environment

## Dataservice Sync
We will also need to ensure that the Data Service studies' short_code, domain, program fields get filled in with the values from study creator. This may need to be part of the data migration code. TBD

gsantia:
Closes #788 

Add three new fields to support new search capabilities on Portal.

`short_code` - A short alphanumeric string that identifies the study in a succinct human friendly way
`domain` - The category of disease being studied
`program` - The administrative organization or group of the study


creator/studies/migrations/0029_add_populate_domain.py
```diff
@@ -0,0 +1,61 @@
...
+        "Birth Defect": "BIRTHDEFECT",
+        "Other": "OTHER",
+        "Cancer": "CANCER",
+        "Cancer,Birth Defect": "CANCERBIRTHDEFECT",
+        "COVID-19": "COVID19",
+    }
+
+    for study in Study.objects.all():
+        kf_id = study.kf_id
+        if kf_id in domains:
```
znatty22, line 33: If a study isn't found in the migration file or doesn't have a domain set, let's set fallback to setting domain to "Unknown"

creator/studies/migrations/0030_add_and_populate_program.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Populate program for all Study objects using the provided data.
+    """
+    df = pd.read_csv(
+        os.path.abspath("creator/studies/migrations/study_metadata.csv")
+    )
+    programs = pd.Series(df.program.values, index=df.kf_id).to_dict()
+    Study = apps.get_model("studies", "Study")
+
+    for study in Study.objects.all():
```
znatty22, line 25: If a study isn't found in the migration file or doesn't have a program listed, let's fallback to  using the study's organization name

creator/studies/migrations/0030_add_and_populate_program.py
```diff
@@ -0,0 +1,58 @@
...
+            study.save()
+
+
+def undo_set_program(apps, schema_editor):
+    """
+    Set program to the Organization name for all Studies.
+    """
+    Study = apps.get_model("studies", "Study")
+    for study in Study.objects.all():
+        if study.organization:
```
znatty22, line 38: I don't think this makes sense here, since this is supposed to undo setting the program. Shouldn't all study programs just be set to null/None?

creator/studies/migrations/0029_add_populate_domain.py
```diff
@@ -0,0 +1,61 @@
...
+            study.save()
+
+
+def undo_set_domain(apps, schema_editor):
+    """
+    Set domain to UNKNOWN for all Study objects.
+    """
+    Study = apps.get_model("studies", "Study")
+    for study in Study.objects.all():
+        study.short_code = "UNKNOWN"
```
znatty22, line 44: Shouldn't we just set this to null/None since its supposed to undo setting the domain?

creator/studies/models.py
```diff
@@ -192,6 +222,14 @@ class Meta:
     def __str__(self):
         return self.kf_id
 
+    def save(self, *args, **kwargs):
```
znatty22, line 225: I don't think this should be here bc this will prevent us from ever being able to set the program explicitly to None (like in the undo function of the migration). This type of logic should probably go in the study create mutation

creator/studies/mutations.py
```diff
@@ -105,6 +108,16 @@ class StudyInput(graphene.InputObjectType):
             "Study version, often the provided by the data access authority"
         )
     )
+    short_code = graphene.String(
```
znatty22, line 111: Should these be required?

creator/studies/mutations.py
```diff
@@ -205,6 +218,15 @@ def mutate(self, info, input, workflows=None):
             )
 
         attributes = sanitize_fields(input)
+
+        # Check short_code is provided and unique
```
znatty22, line 222: If you make this required in the input object, you don't need this check. Graphene will do it for you

znatty22 (changes requested):

creator/studies/mutations.py
gsantia: yes, i forgot to update after we decided they should be

creator/studies/mutations.py
gsantia: it will check it's provided right? but not uniqueness?

creator/studies/mutations.py
gsantia: the input is used for updating studies too, so we can't

creator/studies/migrations/0029_add_populate_domain.py
```diff
@@ -0,0 +1,63 @@
...
+        study.save()
+
+
+def undo_set_domain(apps, schema_editor):
+    """
+    Set domain to UNKNOWN for all Study objects.
+    """
+    Study = apps.get_model("studies", "Study")
+    for study in Study.objects.all():
+        study.short_code = None
```
znatty22, line 46:
```suggestion
        study.domain = None
```

znatty22: Sorry I think I put you on the wrong path with my previous comments on the undo functions. When you rollback these migrations and it calls the undo functions, the rollbacks fail bc domain and program are non-nullable. I *think* that means you were right before when you were setting the values to unknown. And I believe we want the same thing for short_code.

znatty22 (changes requested):

znatty22:
This all looks good but I forgot that we also need another migration that will iterate over all studies in creator, and if the study exists in dataservice, patch it's short_code, domain, and program. I thought that the sync_dataservice job did this, but it actually does the opposite of what we need (updates creator studies with dataservice study data). Let's discuss on slack

creator/studies/migrations/0031_patch_dataservice.py
```diff
@@ -0,0 +1,118 @@
...
+            f"{settings.DATASERVICE_URL}/studies/{kf_id}",
+            json=study_dict,
+            timeout=settings.REQUESTS_TIMEOUT,
+            headers=settings.REQUESTS_HEADERS,
+        )
+    except requests.exceptions.RequestException as e:
+        raise Exception(f"Problem updating study {kf_id}: {e}")
+    
+    # Raise an error if it looks like study failed to update
+    if not resp.status_code == 200 or "results" not in resp.json():
```
znatty22, line 33: I think if a study wasn't found (status 404) in dataservice we probably don't want to raise an exception. Probably just want to log it and return

creator/studies/migrations/0031_patch_dataservice.py
```diff
@@ -0,0 +1,118 @@
...
+        raise Exception(f"Problem updating study {kf_id}: {error}")
+
+
+def patch_attrs(apps, schema_editor):
+    """
+    Patch the short_code, domain, and program attributes of Study objects
+    in the DataService using the provided spreadsheet.
+    """
+    # Do nothing if update DataService feature is not enabled
+    if not (
```
znatty22, line 48: Are you checking this bc these settings are not enabled in dev or something? Wondering why we need this check

creator/studies/migrations/0031_patch_dataservice.py
```diff
@@ -0,0 +1,118 @@
...
+    Study = apps.get_model("studies", "Study")
+    mappings = {
+        "Birth Defect": "BIRTHDEFECT",
+        "Other": "OTHER",
+        "Cancer": "CANCER",
+        "Cancer,Birth Defect": "CANCERBIRTHDEFECT",
+        "COVID-19": "COVID19",
+    }
+
+    for study in Study.objects.all():
```
znatty22, line 70: Since this migration runs last, couldn't you just pull the values you need for short_code, domain, program from the study instance in study creator?

znatty22 (changes requested):

creator/studies/migrations/0031_patch_dataservice.py
gsantia: won't this migration error out anytime someone tries to run study-creator locally without a dataservice up without this check?

creator/studies/migrations/0031_patch_dataservice.py
```diff
@@ -0,0 +1,88 @@
...
+            f"{settings.DATASERVICE_URL}/studies/{kf_id}",
+            json=study_dict,
+            timeout=settings.REQUESTS_TIMEOUT,
+            headers=settings.REQUESTS_HEADERS,
+        )
+    except requests.exceptions.RequestException as e:
+        raise Exception(f"Problem updating study {kf_id}: {e}")
+    
+    # Study not found in DataService
+    if resp.status_code == 400:
```
znatty22, line 34: Not found will be a status code 404

creator/studies/migrations/0031_patch_dataservice.py
```diff
@@ -0,0 +1,91 @@
...
+    Study = apps.get_model("studies", "Study")
+    for study in Study.objects.all():
+        kf_id = study.kf_id
+        study_dict = {
+            "short_code": study.short_code,
+            "domain": study.domain,
+            "program": study.program,
+        }
+        patch_helper(kf_id, study_dict)
+
```
znatty22, line 64: One last request! Can you log the study that was updated along with the study_dict? I'm testing it and I like the warning logs for the ones not found. Thought it would be nice to see what got updated too

znatty22 (changes requested):

creator/studies/migrations/0031_patch_dataservice.py
znatty22: Oh I'm sorry I forgot to thumbs up this. I wanted to double check the value of the feature flag in dev, qa to make sure it was set to true and it is. So I think this is a good check to have

znatty22 (approved):
Looks good 👍 , let's wait to merge until this afternoon

PR #795, opened 2021-10-29
⚗️ Storage analyses sandbox

znatty22:
Placeholder backend for storage analyses app. 

### Create fake storage analyses

image

### Query storage analyses
image

### Schema

image


znatty22:
Add a new metric: files unexpected - files found in S3 but not in upload manifests

PR #801, opened 2021-12-06
✨ Push expected file uploads to Dewrangle

### LINKED ISSUE #800: Push file upload manifests to Dewrangle
Study creator delegates the auditing of files that a user uploads directly to an S3 bucket. It then pulls the results of the audit from Dewrangle so that Data Tracker may populate the S3 accounting dashboard. 

The first step of this is for Study Creator API to push file upload manifests to Dewrangle. This involves parsing the rows in the manifest into FileUploadInvoice objects and sending them in the `fileUploadInvoiceCreate` GraphQL mutation. This process needs to happen:

1. When a user submits a file to Data Tracker which conforms to the File Upload Manifest schema
2. If any files in the above step fail the push to dewrangle, they should be marked as failed and then picked up by a daily job that can retry

znatty22:
Closes #800. **Note** that handling of failures will be implemented in a subsequent PR (#814) 

image


znatty22:
After discussion with @shanewilson 
- Possibly rethink if we want to create study/org that doesn't exist in Dewrangle. Gotta think about sync of entities between two systems (what if someone deletes a study in Dewrangle that Data Tracker knows about, etc)

PR #811, opened 2022-01-07
♻️ Refactor referral token mutation and event messages

XuTheBunny:
# 🚧  🚧  🚧  WORKING IN PROGRESS  🚧  🚧  🚧


creator/events/migrations/0022_add_referral_token.py
```diff
@@ -0,0 +1,25 @@
...
+        ('events', '0021_add_template_download'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='event',
+            name='referral_token',
+            field=models.ForeignKey(blank=True, help_text='Referral token related to this event', null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='events', to='referral_tokens.ReferralToken'),
+        ),
+        migrations.AlterField(
```
znatty22, line 20: This probably shouldn't be in the migration since you only made a change that adds the referral token fk to the event table

creator/referral_tokens/migrations/0003_add_referral_token.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.24 on 2022-01-18 07:27
```
znatty22, line 1: This migration got autogenerated but probably shouldn't be included in this PR since its not related to the changes to add referral token to the event model. Seems like there is something wrong or some missing migration for referral token and organization but we can fix that in a later PR

znatty22 (changes requested):

creator/events/migrations/0022_add_referral_token.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.26 on 2022-01-27 20:22
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('referral_tokens', '0003_add_referral_token'),
```
znatty22, line 10:
```suggestion

```
So the tests are failing bc its trying to setup the test db and can't find the migration `0003_add_referral_token`. Bc you deleted it (as you should have) :)

If you remove this line everything will work

znatty22 (changes requested):

creator/events/migrations/0022_add_referral_token.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.26 on 2022-01-27 20:22
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('events', '0021_add_template_download'),
```
znatty22, line 10:
```suggestion
        # Need to force load the referral_tokens app by making dependency on
        # latest migration. If we don't do this the migration fails bc it
        # cannot find the related model, referral_tokens.ReferralToken
        # https://stackoverflow.com/questions/33496333/valueerror-related-model-uapp-model-cannot-be-resolved
        ('referral_tokens', '__latest__'),
        ('events', '0021_add_template_download'),
```

The latest error you're having is bc django can't find/understand the related model `referral_tokens.Referral_Token`. I found a fix on stack overflow which makes django load the Referral_Token model as a dependency of this migration. See comments above. This should fix the issue

znatty22 (changes requested):

znatty22 (approved):
🗃️

PR #853, opened 2023-08-14
:bug: Sync all studies in dataservice

znatty22:
## Motivation
The API queries the Dataservice when it syncs studies between Dataservice and itself. Unfortunately there is a bug where it only ever requests the first 100 studies since this was maximum # of resources you can fetch in a single request. This is a problem since there are more than 100 studies that need to be synced which causes study search issues on the UI.

## Approach

- Increase the maximum page size from 100 to 1000 in the sync study task

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,3 +14,4 @@ pytest-pycodestyle==2.2.0
 pytest-mock==3.3.0
 pytest-cov==2.12.1
 -e git+https://github.com/codecov/codecov-python.git@4ec70db255cc2dc332fbd01aab3069d9b2e699dc#egg=codecov
+jinja2<3.1.0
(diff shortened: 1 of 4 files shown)


dev-requirements.txt
```diff
@@ -14,3 +14,4 @@ pytest-pycodestyle==2.2.0
 pytest-mock==3.3.0
 pytest-cov==2.12.1
 -e git+https://github.com/codecov/codecov-python.git@4ec70db255cc2dc332fbd01aab3069d9b2e699dc#egg=codecov
+jinja2<3.1.0
```
chris-s-friedman, line 17: Noting that pinning to this version is because 3.1 is the most recent version of jinja, correct?

chris-s-friedman (approved):
Looks good!

